Issue #20, opened 2013-02-21
List vulnerabilities on ruby versions/implementations?


pietro:
Should we list vulnerabilities on ruby versions/implementations? If so, how?

mveytsman:
Closed via #166

PR #45, opened 2013-08-19
[feature] list ruby language vulnerabilities

jeremyolliver:
This is an attempt to implement #20 Feedback is certainly welcome, and needed on the format and file paths I've chosen.

Implementation:

I'm assuming that multiple platforms should be supported e.g. mri, jruby, rbx, and I've borrowed the format largely from the existing one for gems, replacing the `gem` attribute with `platform`

I've also changed the commonly used patch release syntax such as `1.9.3-p447` into `1.9.3.447` this is to make automated parsing by a tool similar to the way [bundler-audit](https://github.com/postmodern/bundler-audit) does for gems, that is, make use of `Gem::Version` (rubygems >= 1.8.0) which correctly handles comparisons using `~>` and other semantic operators on version strings with 4 numeric parts separated by periods.

I've also added almost all of the vulnerabilities listed on http://www.ruby-lang.org/en/security/ into [another feature branch](https://github.com/jeremyolliver/ruby-advisory-db/tree/feature/add-mri-vulnerabilities) which I can send a PR for if this format looks good - but I didn't want this pull request to get too big for now.

### DIFF
--- a/lib/ruby_scrape.rb
+++ b/lib/ruby_scrape.rb
@@ -1,0 +1,87 @@
+require 'rubygems'
+require 'bundler/setup'
+
+require 'pry'
+require 'mechanize'
+require 'yaml'
+require 'date'
+
+class OSVDBRuby
+  attr_accessor :osvdb, :cve, :title, :description, :date, :cvss_v2, :platform, :url, :patched_versions, :page
+  def initialize(url)
+    self.url = url
+    parse!
+  end
+
+  def parse!
+    mech = Mechanize.new
+    self.page = mech.get(url)
+
+    page.search(".show_vuln_table").search("td ul li").each do |li|
+      case li.children[0].text.strip
+      when "CVE ID:"
+        self.cve = li.children[1].text
+      when "Vendor URL:"
+        self.set_platform(li.children[1].text)
+      end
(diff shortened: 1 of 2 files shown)


reedloden:
Any update on this? Would love to have this, and would make a good starting point for trying to fix #165 as well.

phillmv:
All we really need to do is settle on a schema, I think this languished out of inattention rather than lack of interest.

I'd be really down for keeping track of those as well.

reedloden:
The most recent update uses `engine` to represent the interpreter and then uses a more parse-able format of the version (to deal with the crappy -p123 nonsense). I believe those are the only new things introduced here. Any objections to those two things? Otherwise, would be good to get this landed, and we can iterate as needed once we have the first consumer of this data (seems like modifying `bundler-audit` to test this would be a good first step).

phillmv:
OK. I opened up an [new branch](https://github.com/rubysec/ruby-advisory-db/tree/ecosystem-vulns) off master and cherry picked @jeremyolliver's commits. 

I opened another PR in #166 to facilitate collaboration - I can merge things into that branch and edit them directly as opposed to PR commits into master.

PR #90, opened 2014-03-21
Add rack-ssl gem advisory

skorth:
Please double check. Patched version is still missing, i think.


skorth:
@phillmv added patched version & cvss

reedloden:
This was merged in #171. I'll see about adding an advisory for `jruby-rack` itself.

PR #112, opened 2014-11-20
Add unaffected versions for CVE-2014-7818

presidentbeef:


phillmv:
After reviewing this, as I understand this means the proposed PR is right as is (3.0.0 being affected). Thanks @presidentbeef

PR #123, opened 2015-01-26
Add rest-client ruby gem advisory

skorth:


phillmv:
You are awesome, thank you.

jonkessler:
Is this really a vulnerability? Seems pretty dubious to me, especially since we're effectively marking every version of rest-client as vulnerable due to this. Plus, the description on OSVDB is extremely vague.

JonRowe:
Well if all versions of rest-client log passwords in plaintext in the logs they are all "vulnerable', I'd certainly consider this a dubious practise...

jonkessler:
I agree it's dubious. Is it a vulnerability that we need to tell people to not use rest-client for? I don't think so. And they don't just "log passwords in plaintext". If you pass in a password in your url, it logs it. That seems like it's pretty obvious that that would happen, to me.

postmodern:
Is there even a PoC for this? Also where did this advisory originate from?

phillmv:
Agreed re: sparseness of the report details. Weird that no one is credited.

Whether it's a vulnerability… I've written at least one report where we'd mark this down as a "Low" criticality. Logs do leak occasionally. 

It's been forever since I used it, but when included in Rails, does it attach to the Rails logger?

I'll try to make an effort and be a little more critical henceforth.

skorth:
FYI https://github.com/rest-client/rest-client/pull/352

PR #124, opened 2015-02-05
Add ruby-saml Gem advisory

skorth:
https://github.com/onelogin/ruby-saml/blob/master/changelog.md#082-jan-26-2015


phillmv:
Thank you sir.

PR #127, opened 2015-02-26
Add xaviershay-dm-rails Gem advisory

skorth:
I just added xaviershay-dm-rails cause of the osvdb entry.
POC: http://www.vapid.dhs.org/advisory.php?v=115

@xaviershay FYI

### DIFF
--- a/gems/xaviershay-dm-rails/OSVDB-118579.yml
+++ b/gems/xaviershay-dm-rails/OSVDB-118579.yml
@@ -1,6 +1,7 @@
 ---
 gem: xaviershay-dm-rails
 osvdb: 118579
+cve: 2015-2179
 url: http://osvdb.org/show/osvdb/118579
 title: |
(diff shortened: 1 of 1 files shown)


xaviershay:
fwiw the gem has been yanked. No idea why it was published in the first place...

phillmv:
Hah. Is it even still available for download then? The page on rubygems lists a dependency on rails 3.0 beta, so trololol.

xaviershay:
argh there are old versions too. I just need to blow this thing out of existence. Gonna yank all the versions.

skorth:
@xaviershay thx ;-)

"This gem has been yanked, but it is still available for download for other gems that may have depended on it". I think it still makes sense to add the advisory, or?

skorth:
@phillmv i added the cve-id

phillmv:
Eh, alrighty!

Thanks, @skorth and @xaviershay

PR #128, opened 2015-02-28
Add Doorkeeper ruby gem advisory

skorth:
Maybe someone can double check if the 'patched_versions' fits.

https://github.com/doorkeeper-gem/doorkeeper/issues/576
https://github.com/doorkeeper-gem/doorkeeper/blob/master/CHANGELOG.md#212


postmodern:
Looks good.

PR #135, opened 2015-03-22
Add Ruby kompanee-recipes Gem advisory

skorth:
Just added it cause of an older osvdb entry...for the sake of completeness.

https://rubygems.org/gems/kompanee-recipes
http://osvdb.org/show/osvdb/108593
http://www.vapid.dhs.org/advisory.php?v=67


phillmv:
Ah, you can always tell a Larry Cashdollar entry.

Thanks!

PR #137, opened 2015-03-25
Add rest-client ruby gem advisory

skorth:
https://github.com/rest-client/rest-client/blob/master/history.md#180


phillmv:
Sorry @skorth I think the language in #138 makes a bit more sense. <3

skorth:
:+1:

PR #138, opened 2015-03-25
Add rest-client CVE-2015-1820.

ab:
https://github.com/rest-client/rest-client/issues/369

### DIFF
--- a/gems/rest-client/CVE-2015-1820.yml
+++ b/gems/rest-client/CVE-2015-1820.yml
@@ -1,0 +1,21 @@
+---
+gem: rest-client
+cve: 2015-1820
+osvdb: 119878
+url: https://github.com/rest-client/rest-client/issues/369
+title: 'rubygem-rest-client: session fixation vulnerability via Set-Cookie headers in 30x redirection responses'
+date: 2015-03-24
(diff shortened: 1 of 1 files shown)


skorth:
Same as https://github.com/rubysec/ruby-advisory-db/pull/137 ?

ab:
Yeah, but the OSVDB entry has some very poor, confusing language around session fixation attacks.

phillmv:
@ab awesome. Quick q, any idea on when the mitre page will update? The github issue seems to be the most informative atm, maybe we should link to that instead.

ab:
MITRE's pretty backlogged, so not sure when they'll have content up. Happy to have the link point wherever. I wasn't sure if you generally preferred links to point to the official CVE pages or to an upstream bug tracker.

phillmv:
@ab Cool. I did some soul searching and I think the *best* source of info ought to be in the URL. That Github issue works great, and if you rly wanna look up MITRE or OSVDB we've included the ids.

Give'er an update real quick and let's call merge it in. 

Thanks!

phillmv:
Awesome. Thanks @ab, @skorth

PR #141, opened 2015-03-26
Add Fat Free CRM Ruby Gem advisories

skorth:
Please double check ;-)


phillmv:
Thanks, as always :D.

PR #142, opened 2015-03-27
Update Gems with CVE numbers

VanessaHenderson:
- [ ] Update various gems with CVE numbers
- [ ] Update version range on rest-client gem based on GitHub issue


phillmv:
Awesome! Thank you.

For future reference, note that we support an 'unaffected_version' field to distinguish between versions that are patched and versions that are not vulnerable - i.e.1b32d551936e811488f2264dddf882ec00d1b8ce

PR #147, opened 2015-05-13
Add open-uri-cached ruby gem

skorth:
Didn't found a poc or any other additional information.

### DIFF
--- a/gems/open-uri-cached/OSVDB-121701.yml
+++ b/gems/open-uri-cached/OSVDB-121701.yml
@@ -1,0 +1,13 @@
+---
+gem: open-uri-cached
+cve: 2015-3649
+osvdb: 121701
+url: http://osvdb.org/show/osvdb/121701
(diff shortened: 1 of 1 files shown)


skorth:
FYI @mscherer

phillmv:
Cool. The lib was last updated 3 years ago so not holding my breath on patched versions coming out.

mscherer:
I posted a bit more information on http://seclists.org/oss-sec/2015/q2/373 , and I can also explain a bit the issue if needed without trouble ( but I would prefer to have some guidance on what to explain, since to me, it look like a traditional /tmp issue ( cf https://cwe.mitre.org/data/definitions/377.html for example ) so I am not sure what to tell more ).

skorth:
@mscherer thanks a lot for your reply, i added the new url https://github.com/rubysec/ruby-advisory-db/pull/148 to the advisory.

PR #149, opened 2015-05-17
Adding missing vulns compared to bugzilla.redhat.com

jrusnack:
I compared ruby-advisory-db with bugzilla.redhat.com and added a few missing vulnerabilities. There are more, but I didn`t want to go further back beyond 2011. 

Most of the time I took information from OSVDB as primary source, only in cases where this info was missing/incomplete I used bugzilla.redhat.com.


skorth:
Kudos! :+1:

VanessaHenderson:
the CVSS score for [this item] (https://github.com/jrusnack/ruby-advisory-db/blob/bugzilla-cve/gems/ldap_fluff/OSVDB-90579.yml) is wrong, its 4.3 not 5 according to NVD anyway

jrusnack:
@VanessaHenderson  (disclaimer: I work at RH Product Security) For this one I went with what`s in bugzilla, as redhat is upstream for this gem and NVD would surely have less info about the flaw. After your comment I checked, the details on how to reproduce are in a private bug 871635. 

So the CVSS score difference is in Access Complexity: Medium or Low. Definitions from the nvd (https://nvd.nist.gov/cvss.cfm?calculator&version=2) say low when no specialized access conditions exist. Medium when:
* The attacking party is limited to a group of systems or users at some level of authorization, possibly untrusted. Not applicable here, no authorization before one tries to log in.
* Some information must be gathered before a successful attack can be launched. Not applicable.
* The affected configuration is non-default, and is not commonly configured. Again not applicable, MS AD is prevalent in enterprise environments.
* The attack requires a small amount of social engineering... Not applicable.

And I think the above is clear even from the short description in https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2012-5604. No idea why NVD says AC:M.

@postmodern I did redcloth rather than RedCloth because I`ve seen other gems spelled like that - for example VladTheEnterprising is vladtheenterprising. I don`t have preference either way - we can change both to use capitalization.

@VanessaHenderson @postmodern Thanks for review !!

postmodern:
Gem names are case sensitive, and the RedCloth gem is indeed capitalized
(https://rubygems.org/gems/RedCloth). The gem directory names should
match the canonical gem names. This allows bundler-audit to match
ruby-advisory-db advisories against installed gems.

RE CVSS, I would send an email to moderators@osvdb.org. It might just be
a case of OSVDB having outdated data.

jrusnack:
@postmodern I renamed three directories that I know did not match the names. 

For CVSS I sent the mail.

Thanks!

phillmv:
@jrusnack @skorth @VanessaHenderson you are all wonderful

Jan, you rock, thanks. Seeing as everything seems to have been addressed, I'm calling this done

postmodern:
:+1: thank you for the added coverage!

PR #150, opened 2015-06-07
Added bson, moped data injection CVE-2015-441{0,2}

phillmv:
@postmodern 
I added a `vendor_patch` attribute. Let me know if you disagree. I was stuck between `vendor_uri` or `vendor_context`.

Also, @durran and @estolfo, sorry for the extra work but - the purpose of these files is to ensure people can upgrade to safe versions. 

Could you confirm if the version information is correct?

Much obliged,

### DIFF
--- a/gems/bson/CVE-2015-4412.yml
+++ b/gems/bson/CVE-2015-4412.yml
@@ -1,0 +1,16 @@
+---
+gem: bson
+cve: 2015-4412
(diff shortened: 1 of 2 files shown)


durran:
@phillmv Yes the patched version numbers are correct for both `bson` and `moped`. Thanks!

adrianmacneil:
:+1:

PR #161, opened 2015-07-01
Update CVE-2015-3226.yml

habermann24:
Rails 3.2.x was not affected by this CVE

See:
https://groups.google.com/forum/#!topic/rubyonrails-core/qBUqVlXERag

### DIFF
--- a/gems/activesupport/CVE-2015-3226.yml
+++ b/gems/activesupport/CVE-2015-3226.yml
@@ -45,6 +45,7 @@ description: |
     end 
 
 unaffected_versions:
+  - "~> 3.2.22"
   - "~> 4.0.0"
 
 patched_versions:


phillmv:
Hi!

So, this got addressed by @jiop in #158 but we merged it into the wrong branch.

Thanks for catching this, but note the semantic difference between your commit and Jeremie's - the email you linked notes that "Rails < 4.1 is not affected." but your commit is singling out 3.2.22 and up as being unaffected.

In lieu of this I will be closing this PR and merging in #162 instead.

Sorry about that but thanks for the effort nonetheless! You allowed us to catch an error :D.

habermann24:
Oops. you're right, sorry. Great that it's fixed now, thanks :smile:

PR #166, opened 2015-07-15
Ecosystem vulns

phillmv:
New spot for discussion, patches prior to fully adding it to master.

Benefit of this over #45 is we can accept other PRs into this branch prior to merging with master, and collect a bunch before committing to a schema.


reedloden:
Can you import everything from https://github.com/jeremyolliver/ruby-advisory-db/tree/feature/add-mri-vulnerabilities to this new branch as well?

phillmv:
Done.

So one thing, do we distinguish rubygems as a separate entity? Or could we repurpose "engine" somehow?

reedloden:
Are there other things that fit into the category that rubygems is? I mean, it's Ruby's package manager, but are there other things that are bundled with Ruby? Things coming to mind are `rake` and possibly the bindings type stuff (but I guess the latter would count as part of Ruby directly).

reedloden:
Oh, `rake` is a gem itself. See the things under https://github.com/ruby/ruby/tree/trunk/lib. Are any of those besides `rubygems` managed as a separate project? We could possibly overload `engine` or just create a new `library` class.

reedloden:
We also lost `lib/ruby_scrape.rb` in this new branch, but I think we should just modify the existing `lib/scrape.rb` to support rubies rather than duplicating a lot of code.

jeremyolliver:
Glad to see some more activity on this :) The timing is uncanny. I literally published an internal security report for my company earlier today, in which I lamented I was not able to easily cover how many vulnerabilities each ruby-lang version had, so I just scoped the report to gems instead.

I'd be happy to help out, and contribute, as I'd really like to see this become a thing. I also found today some extra sources (http://www.cvedetails.com/version-list/7252/12215/1/Ruby-lang-Ruby.html), which may or may not be easier to categorize existing CVE information from, than the existing ruby-lang security page. I don't think it's an appropriate target for scraping, but it's at least easier to read for humans than other places I've seen.

jeremyolliver:
In regards to other categories such as libs - rubygems of course stands out, though there are some oddities, with some things in stdlib like minitest (can't think of other examples right now) which are both packaged with ruby, and upgradeable as a gem. I expect any vulnerabilities in stdlib are going to be issued as CVE against a ruby-lang version - even if you can workaround by installing a newer gem instead of upgrading ruby.

Having said that though, I'd prioritize having a working set of ruby-lang vulnerabilities before worrying about that.

reedloden:
@jeremyolliver, anymore advisories you can create would be helpful. It's been a few years since you last did a scrape, so should be several more at least. :)

jeremyolliver:
I'd be happy to help out with some of that effort - I won't add it to this PR though, I'd rather this initial batch get merged first.

postmodern:
:+1: for putting them in `rubies`. I would however use `ruby` instead of `mri`, to match `RUBY_ENGINE` under MRI.

As for scraping, maybe we could convince [www.ruby-lang.org](https://github.com/ruby/www.ruby-lang.org) to use a jekyll template for their CVE news postings? This way the CVE could be stored in the YAML metadata.

dwradcliffe:
ping @zzak RE: ruby-lang.org

postmodern:
@dwradcliffe I'll submit an issue over there and let them ponder it.

phillmv:
OK. I added some tests. I can't think of anything missing, merging in!

phillmv:
(I recommend waiting just a bit before setting tooling support in stone)

jeremyolliver:
Fantastic work - thanks for this :)

PR #167, opened 2015-07-15
Add Travis CI support for automagically updating rubysec website

### LINKED ISSUE #160: Update rubysec.github.io on every commit
Should use Travis CI / Circle CI to automagically update the `rubysec.github.io` repository and push changes live as new advisories are added to the `ruby-advisory-db` repository.

Right now, http://rubysec.com is woefully out of date. :(

Happy to help here if I can...

reedloden:
Update advisories on rubysec website every time a successful commit
is made on the master branch.

Fixes #160.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: ruby
+
+sudo: false
+
+cache: bundler
+
+deploy:
+  provider: script
+  script: scripts/post-advisories.sh
+  on:
+    branch: master
+
+notifications:
+  irc: chat.freenode.net#rubysec
+
+env:
+  global:
+    - secure: "o/aIfTQb1n8YP+Mq3be3XQfcENNmwHtjWcO2pTz6b6uQqsg1p+vPhhGQM7FJXID1rU5Nyp9hwwU/QpQAU99mx/hdAxqSl1RxYj6vknrngk00KpzC0K+1qQMuM+NFNa09FmJTn8KkC1GjNrSCR/U0o946HK+svk/aL4a05zRAA3Y="
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 Gemfile.lock
+_site
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
+gem 'rake'
 gem 'rspec'
--- a/scripts/post-advisories.sh
+++ b/scripts/post-advisories.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -o errexit -o nounset
+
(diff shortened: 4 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,24 @@
+language: ruby
+
+rvm:
+- 2.2
+- 2.1
+- 2.0.0
+- 1.9.3
+- 1.8.7
+- jruby-19mode
+- rbx-2
```
dwradcliffe: Probably just need one ruby version here

reedloden: Happy to keep it as one, but figured it would be good to ensure the database can be parsed/consumed by various versions to ensure compatibility. However, perhaps that's more a job for bundler-audit...

dwradcliffe: I think this will try to update the website 7 times.

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+
+sudo: false
+
+cache: bundler
+
+after_success:
```
dwradcliffe: Could/should we use the `deploy` phase here? http://docs.travis-ci.com/user/customizing-the-build/#Deploying-your-Code

dwradcliffe:
It would be great if we could use the rake tasks in the website repo directly, instead of having the code in both repos. Some changes to the other repo might be necessary but I think it would be worth it.

.travis.yml
reedloden: Hah, you might be right there... I'll look into that to see if there's a way to just do it once, but if it doesn't work, I'll just strip out the versions and leave the default.

.travis.yml
reedloden: Sadly, no. I looked into that originally. The only `deploy` step for GitHub supported by Travis CI is for actual releases (tags, tarballs, etc.). See http://docs.travis-ci.com/user/deployment/ for what they support, and http://docs.travis-ci.com/user/deployment/releases/ for the only GitHub-related deployment step.

dwradcliffe: What about http://docs.travis-ci.com/user/deployment/script/?

reedloden:
Well, I was going to remove it from the website repo (as it isn't needed then), but since I'm checking out the repo, I guess I can keep stuff in there. Will refactor.

dwradcliffe:
:+1: Good to be able to run it manually too.

.travis.yml
jeremyolliver: You could put a bash conditional on the after_success script to only run based on the current ruby version matching just one of these - but given that this repo is primarily data, rather than a code library, I think just one ruby version here would be fine

reedloden:
Ok, redid how this works. https://github.com/rubysec/rubysec.github.io/pull/10 will need to be landed before this.

PR #172, opened 2015-07-20
Add OSVDB-115891 for activeresource

reedloden:
I'm adding this here for now so it doesn't get forgotten, but I ran across http://osvdb.org/show/osvdb/115891 being mentioned in https://github.com/thesp0nge/dawnscanner/issues/112, which talks about some unspecified activeresource vulnerability.

After lots of digging, I found https://github.com/rails/activeresource/issues/143 (and subsequent PR to fix in https://github.com/rails/activeresource/pull/144). However, doesn't look like the PR ever landed, so this still might be a real issue? I've left patched_versions blank for now.

Since the OSVDB entry and various GitHub issue/PRs are all public, keeping this public as well, but would be good to involve activeresource devs (rails devs, I guess) to see if this is a real issue or not. I'll poke security@rubyonrails about this.

### DIFF
--- a/gems/activeresource/OSVDB-115891.yml
+++ b/gems/activeresource/OSVDB-115891.yml
@@ -1,0 +1,8 @@
+--- 
+gem: activeresource
+framework: rails
+osvdb: 115891
+url: http://www.osvdb.org/show/osvdb/115891
+title: Active Resource (ARes) Gem for Ruby lib/active_resource/base.rb Thread Object Instantiation Unspecified Issue
+date: 2014-09-24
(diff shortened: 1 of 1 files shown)


postmodern:
Definitely double check with Rails Core, so we don't wrongly flag every version of activeresource.

reedloden:
Yup, @tenderlove is already looking into this.

reedloden:
@tenderlove @rafaelfranca @jeremy any thoughts here? The OSVDB link is returning 404 now, so just see the 'Files changed' tab to see what it actually said.

jeremy:
@reedloden This was triaged as a bugfix, not a vuln. Mentioned you in the relevant HackerOne report.

PR #176, opened 2015-07-20
Audit of @lcashdol's ruby gem OSVDB reports to ensure we are tracking all of them

reedloden:
I went through http://osvdb.org/creditees/6795-larry-w-cashdollar and tried to make sure we were tracking all the ruby gems @lcashdol has reported issues for (or if we had them, they were updated with the right information)... There's probably some I missed, but this PR covers most of them, I think.


gems/codders-dataset/OSVDB-108582.yml
```diff
@@ -0,0 +1,8 @@
+---
```
VanessaHenderson: Should this not be combined into 1 with OSVDB-108583 as they are the same CVE? Opinion?

reedloden: They are separate in OSVDB, though. MITRE treated postgresql and mysql the same. OSVDB split them into separate issues. I prefer we track all OSVDB and CVE IDs for Ruby issues so there's never a question as to whether we have something in our database. It would be nice if we had our own numbering system and then could attach CVEs and OSVDB IDs to those issues, but that's out-of-scope here.

phillmv: Hrm. I think the distinction to be made is more that we should have a copy per flaw-people-should-be-aware-of. MITRE treating both mysql and postgres flaws as the *same thing*, from the user's perspective, is correct imho. 

I think we should be *exhaustive* in terms of having sources for things, but meh on having exact copies from upstream.

phillmv: This is to say I'm more in favour of just the CVE approach at least for this one.

gems/fastreader/OSVDB-91232.yml
```diff
@@ -1,12 +1,13 @@
 ---
 gem: fastreader
-cve: 2013-1876
+cve: 2013-2615
```
phillmv: good catch

gems/codders-dataset/OSVDB-108582.yml
reedloden: Then we should support tracking multiple IDs per report. As in, I'm fine with their being only one "vulnerability", but I want to track the multiple OSVDB IDs somehow at least, as otherwise, it's a pain later when matching up OSVDB's DB to ours.

phillmv:
God bless @lcashdol, looks good and merging in.

PR #177, opened 2015-07-20
More auditing of ruby gem OSVDB reports to ensure we are tracking all of them

reedloden:
Went through http://osvdb.org/search?kthx=search&page=1&search[referencetypes]=&search[refid]=&search[text_type]=titles&search[vuln_title]=gem (all 3 pages) to make sure we were tracking all the ruby gems... Still needs another run through once all my PRs have been merged.


phillmv:
killer job, mate.

PR #179, opened 2015-07-20
Adding OSVDB-121500 for i18n

reedloden:
**NOT READY FOR MERGING**

http://osvdb.org/show/osvdb/121500

This refers to https://github.com/svenfuchs/i18n/pull/289, which hasn't yet made it to a released version of `i18n`. Perhaps somebody could poke them to get that done so we can add the advisory to the db.


postmodern:
Oops, merged too soon :(

PR #183, opened 2015-07-21
rack-ssl 1.3.4 is also patched for this issue

rafaelfranca:
See https://github.com/josh/rack-ssl/pull/34 and https://github.com/josh/rack-ssl/compare/v1.3.3...v1.3.4

### DIFF
--- a/gems/rack-ssl/OSVDB-104734.yml
+++ b/gems/rack-ssl/OSVDB-104734.yml
@@ -8,4 +8,4 @@ date: 2013-07-09
 description: rack-ssl Gem for Ruby contains a flaw that allows a reflected cross-site scripting (XSS) attack. This flaw exists because the program does not validate input passed via error messages before returning it to users. This may allow a context-dependent attacker to create a specially crafted request that would execute arbitrary script code in a user's browser session within the trust relationship between their browser and the server.
 cvss_v2: 4.3
 patched_versions:
-  - ">= 1.4.0"
+  - ">= 1.3.4"


reedloden:
That's not the security issue discussed.

https://github.com/josh/rack-ssl/commit/9d7d7300b907e496db68d89d07fbc2e0df0b487b is the fix, and that's only in 1.4.0 and greater.

reedloden:
There's also a few follow-up diffs if you did want to backport that...

pedelman:
Adding this in since it is a related discussion.

https://github.com/rails/rails/issues/20964

rafaelfranca:
It is the same fix but with a different patch. The patch present on 1.4.x is rescuing the exception, the patch on 1.3.4 just stopped to parse the URI that means that the exception is not raised anymore and in consequence the resulting exception is not displayed (the reason for the XSS).

reedloden:
Ah, cool. That wasn't clear at all when I was reading through commits when writing this advisory. Didn't seem to be any link at all.

rafaelfranca:
I remember I asked Josh to backport the commit to 1.3.x and we end up with that patch. But the issue seems to be lost when he disabled the issue tracker of the project.

phillmv:
OK, from one Mendonça to another thanks @rafaelfranca

smlsml:
It sounds like this is fixed/avoided in 1.3.4 but it still reports as a problem due to this:

https://github.com/rubysec/ruby-advisory-db/blob/970b2377902c5d034a227c4d422ad691f95262e0/gems/rack-ssl/OSVDB-104734.yml

How can we get that updated so it stops warning us?

Issue #164, opened 2015-07-12
Merge the two sup vulnerabilities


mveytsman:
While https://github.com/rubysec/ruby-advisory-db/tree/master/gems/sup have different CVEs, it's the same vuln with the same fixed versions

Maybe we should merge them?

jeremyolliver:
If you check the CVE descriptions - one is for RCE via the filename, the second is for RCE via the content_type, so they are slightly different

reedloden:
Yeah, rather keep CVEs separate. Just ran into this with #168. Saves a lot of historical work later by tracking them separately.

mveytsman:
Thinking about this more, I disagree. 

We should have a schema to track other CVEs that a vulnerability refers to, but no reason to have 2 separate rubysec vulns for this. It only leads to confusion, since in our database they look like duplicates, and no one gains any new information by seeing the two entries.

The schema should look like:

```yaml
other_cves: 
 - 2013-4478
```

reedloden:
`other_cves` seems weird (which is the primary and which goes in other?)... any reason why not to permit `cve` to have multiple ones? In that case, what do you name the file?

mveytsman:
Agreed, it should be `cve`.

For now, for this particular file we can stick with the osvdb number. 

It is high time we pick something else for file naming.

postmodern:
@jasnow has added the `related:` CVE and GHSA IDs to the two vulnerabilities. We should avoid merging CVEs as they are supposed to be individually enumerated vulnerabilities.

PR #189, opened 2015-07-28
Use actual advisory URL for sup vulnerabilties

reedloden:
Think this just makes more sense, and it has way more information than the OSVDB entry.


mveytsman:
:+1: 

I do think these should be merged (https://github.com/rubysec/ruby-advisory-db/issues/164)

mveytsman:
heh merged was an overloaded term above :)

PR #191, opened 2015-07-29
Add CVE-2015-1840 for jquery-ujs and modify advisory for jquery-rails

reedloden:


gems/jquery-rails/CVE-2015-1840.yml
```diff
@@ -1,11 +1,9 @@
 ---
 gem: jquery-rails
 cve: 2015-1840
-url: https://groups.google.com/forum/#!topic/ruby-security-ann/XIZPbobuwaY
-title: |
-  CSRF Vulnerability in jquery-ujs and jquery-rails
+url: http://blog.honeybadger.io/understanding-the-rails-jquery-csrf-vulnerability-cve-2015-1840/
```
phillmv: OK, so - url I feel should refer to canonical upstream disclosure.

I definitely like the idea of linking explanatory posts, though. Can you think of an acceptable schema addition?

reedloden: Could just support multiple urls, with the first one treated as the authoritative. Simple but useful schema change.

reedloden: As in:
```
url:
  - http://foo.bar
  - http://baz.blah
```

reedloden: Otherwise, a separate list of `references` or `related_urls` would work.

reedloden: To unblock this, I just added the URL to the description and swapped back to the old URL.

phillmv:
unfortunately comments get hidden when diffs change, so you wrote 

>Otherwise, a separate list of `references` or `related_urls` would work.

That is exactly what I had in mind. I think references as an array type would be good. For now let's merge this in and create an issue for that.

PR #195, opened 2015-08-02
Add / modify advisories for spree

reedloden:
Add a bunch of advisories for `spree`, as well as modify two existing ones.

Note that `spree` is a bit weird in that it's really split into a bunch of different gems. These gems all keep the same version for the release. Some of this fixes cross these sub-gem boundaries, so hard to keep track of things. I made the decision to just keep everything tied to the `spree` gem itself, even though some other sub-gems are affected.

### DIFF
--- a/gems/spree/OSVDB-73751.yml
+++ b/gems/spree/OSVDB-73751.yml
@@ -1,0 +1,11 @@
+---
+gem: spree
+osvdb: 73751
+url: https://spreecommerce.com/blog/security-fixes
+title: Spree Content Controller Unspecified Arbitrary File Disclosure
+date: 2011-04-19
+description: |
+  Spree Gem for Ruby would allow a user to request a specially crafted URL and
+  expose arbitrary files on the server
+patched_versions:
+  - ">= 0.50.1"
--- a/gems/spree/OSVDB-90865.yml
+++ b/gems/spree/OSVDB-90865.yml
@@ -1,0 +1,20 @@
+---
+gem: spree
+cve: 2013-2506
+osvdb: 90865
+url: https://spreecommerce.com/blog/multiple-security-vulnerabilities-fixed
+title: |
+  Spree app/models/spree/user.rb Mass Role Assignment Remote Privilege
+  Escalation
+date: 2013-02-21
+description: |
+  Spree contains a flaw that leads to unauthorized privileges being gained. The
+  issue is triggered as certain input related to mass role assignment in
+  app/models/spree/user.rb is not properly verified before being used to update
+  a user. This may allow a remote attacker to assign arbitrary roles and gain
+  elevated administrative privileges.
+cvss_v2: 4.0
+patched_versions:
+  - ~> 1.1.6
+  - ~> 1.2.0
+  - ">= 1.3.0"
--- a/gems/spree/OSVDB-91217.yml
+++ b/gems/spree/OSVDB-91217.yml
@@ -1,0 +1,17 @@
+---
+gem: spree
+cve: 2013-1656
+osvdb: 91217
+url: https://spreecommerce.com/blog/multiple-security-vulnerabilities-fixed
+title: |
+  Spree payment_methods_controller.rb payment_method Parameter Arbitrary Ruby
+  Object Instantiation Command Execution
+date: 2013-02-21
+description: |
+  Spree contains a flaw that is triggered when handling input passed via the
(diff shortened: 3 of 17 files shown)


gems/spree/OSVDB-73751.yml
```diff
@@ -0,0 +1,11 @@
+---
+gem: spree
+osvdb: 73751
```
VanessaHenderson: The description on this OSVDB entry is non-existent at all, when they've got the updates working again are you going to submit a new description for the item?

reedloden: Yup, as soon as OSVDB is back to working order, I will go back and update all of this. Main thing is to get them added now, and the URL provides the real description.

gems/spree/OSVDB-90865.yml
```diff
@@ -0,0 +1,20 @@
+---
+gem: spree
```
VanessaHenderson: This is also related to the spree_auth_devise gem.  Here is the [GitHub commit ] for the spree_auth_devise gem which is the same as the spree, another advisory might need to be added for that gem  ( https://github.com/spree/spree_auth_devise/commit/038d74771d3b5c13d13b738b73dfda1033a99f65 )with the fix

VanessaHenderson: Also after examining the Source Code it would appear that spree_auth before 1.1.6 (which was never officially released) is also vulnerable. Its my understanding that 1.2.0 and 1.3.0 aren't actually vulnerable. The said file doesn't exist in these versions

gems/spree/OSVDB-91217.yml
```diff
@@ -0,0 +1,17 @@
...
+  Object Instantiation Command Execution
+date: 2013-02-21
+description: |
+  Spree contains a flaw that is triggered when handling input passed via the
+  'payment_method' parameter to payment_methods_controller.rb. This may allow
+  a remote authenticated attacker to instantiate arbitrary Ruby objects and
+  potentially execute arbitrary commands.
+cvss_v2: 4.3
+patched_versions:
+  - ">= 2.0.0"
```
VanessaHenderson: The CVE says the affected versions are 1.0.x-1.3.2 so wouldn't the patched version be 1.3.3? Or did you research more and find it was 2.0.0?

postmodern: Looks like [spree 1.3.3](https://rubygems.org/gems/spree/versions) was released June 13, 2013. So probably should be `">= 1.3.3"`.

gems/spree/OSVDB-90865.yml
reedloden: https://rubygems.org/gems/spree_auth was released, sadly...

gems/spree/OSVDB-91217.yml
reedloden: The patches were never backported to old branches, so >= 2.0.0 is correct.

https://github.com/spree/spree/commits/1-3-stable/core/app/controllers/spree/admin/payment_methods_controller.rb

gems/spree/OSVDB-90865.yml
reedloden: I added copies for spree_auth and spree_auth_devise as well.

reedloden:
Added a few more advisories as well, thanks to OSVDB providing assignments.

gems/spree_auth/OSVDB-90865.yml
```diff
@@ -0,0 +1,18 @@
...
+date: 2013-02-21
+description: |
+  Spree contains a flaw that leads to unauthorized privileges being gained. The
+  issue is triggered as certain input related to mass role assignment in
+  app/models/spree/user.rb is not properly verified before being used to update
+  a user. This may allow a remote attacker to assign arbitrary roles and gain
+  elevated administrative privileges.
+cvss_v2: 4.0
+patched_versions:
+  - ">= 1.1.6"
```
VanessaHenderson: This version was never released so I'm not sure that its worth putting it in

reedloden: Prefer to leaving `patched_versions` blank here, meaning all known versions are vulnerable?

phillmv: Yeaaaah, let's err on the side of "unreleased patched version" means everything is still vulnerable.

phillmv:
@reedloden thank you for the submissions!

@VanessaHenderson thank you for double checking!

PR #196, opened 2015-08-02
Add OSVDB-125675, OSVDB-125676, OSVDB-125677, and OSVDB-125678 for sidekiq

reedloden:
All affecting Sidekiq::Web

### DIFF
--- a/gems/sidekiq/OSVDB-125675.yml
+++ b/gems/sidekiq/OSVDB-125675.yml
@@ -1,0 +1,9 @@
+---
+gem: sidekiq
+osvdb: 125675
+url: https://github.com/mperham/sidekiq/pull/2422
+title: Sidekiq Gem for Ruby Multiple Unspecified CSRF
+date: 2015-07-06
+description: Sidekiq::Web lacks CSRF protection
+patched_versions:
+  - ">= 3.4.2"
--- a/gems/sidekiq/OSVDB-125676.yml
+++ b/gems/sidekiq/OSVDB-125676.yml
@@ -1,0 +1,11 @@
+---
+gem: sidekiq
+osvdb: 125676
+url: https://github.com/mperham/sidekiq/issues/2330
+title: |
+  Sidekiq Gem for Ruby web/views/queue.erb CurrentMessagesInQueue Element
+  Reflected XSS
+date: 2015-06-04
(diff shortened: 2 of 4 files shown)


reedloden:
Descriptions aren't super-detailed, but I'll go back and add more detailed ones later. The link to the PRs/issues should be enough for most people.

phillmv:
Hrm, this advisory is making a great argument for settling for some alternate OSVDB id schema.

How about related_osvdb that takes a list?

reedloden:
you mean like tying different OSVDB entries together? Only two of the four relate to the same issue.

but yeah... I worry that `related_osvdb` is too specific... I would like to have a more generic `references` that can have sub sections for other things, which could include other related OSVDBs.

phillmv:
So my caveat with a generic `references` vs a specific `related_attribute_name_here`:

The specific schema will be easier to test, implement and change later if need be. We should be specific first and then abstract as we collect more uses, is my feeling.

Does that make sense?

reedloden:
Fair enough... Can agree with that for a generic `references` thing, but what about just `related`?

Does it make sense to do:
``` lang=yaml
osvdb_related:
  - 12345
  - 67890

cve_related:
  - 2015-1234
  - 2014-5678
```

OR

``` lang=yaml
related:
  osvdb:
    - 12345
    - 67890

  cve:
    - 2015-1234
    - 2014-5678
```

phillmv:
@reedloden settled for the more generic related :).

reedloden:
Eh, I wish we kept the actual advisory in the case of multiple ones and relied on the consumers of the DB to ignore related ones... Otherwise, we're doing the users of our DB a disservice, as they may feel they have mitigated some issue, while the 'related' one still affects them.

phillmv:
@reedloden in this case in particular, the only different bit of information that was different was the osvdb id.

If the title, description, url and patched version is identical then imho (full disclosure:  I'm biased as I maintain a downstream consumer of this db) all use cases are better off. We maintain "referential completeness" in terms of sourcing data and everyone knows what to do.

I think in future situations where there are two different bugs, and as a result there are two distinct things to mitigate then that makes total sense to keep them as separate. Otherwise it's just noise.

Does that make sense?

PR #199, opened 2015-08-07
Adding CVE-2014-0156

jrusnack:
This is a PR to include CVE-2014-0156 OS command injection vulnerability in awesome_spawn.

The vulnerability was found by Jan Rusnacko, reported directly to upstream maintainer on 2014-03-28, CVE ID was assigned by Red Hat and the issue was fixed in commit https://github.com/ManageIQ/awesome_spawn/commit/e524f85f1c6e292ef7d117d7818521307ac269ff

### DIFF
--- a/gems/awesome_spawn/CVE-2014-0156.yml
+++ b/gems/awesome_spawn/CVE-2014-0156.yml
@@ -1,6 +1,6 @@
 ---
 gem: awesome_spawn
-cve: CVE-2014-0156
+cve: 2014-0156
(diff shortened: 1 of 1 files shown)


phillmv:
Cool. Thanks, Jan!

As an aside, any reason why this didn't make it out to MITRE?

jrusnack:
Good question. I found this as part of internal audit, but did not handle, so I can only guess why this was let to slip through.

PR #200, opened 2015-08-09
Reedloden's renamed gems

phillmv:


phillmv:
@reedloden created separate pull request just to document my changes,

I thought about it and I think it makes more sense to keep the old patched_version info in there, while adding a note that the gem has been renamed.

While we're at it, no patch for https://github.com/rubysec/ruby-advisory-db/blob/reedloden-renamed-gems/gems/fog-dragonfly/OSVDB-97854.yml ?

reedloden:
@phillmv No patch in `fog-dragonfly`. The patch is in `dragonfly` (the renamed version). https://github.com/rubysec/ruby-advisory-db/blob/master/gems/dragonfly/OSVDB-97854.yml

PR #203, opened 2015-09-17
Add CVE-2015-7225 to devise-two-factor

f3ndot:
Awaiting identifier from OSVDB.

Was just assigned the CVE identifier, so it's not on MITRE's site yet.


f3ndot:
For the lazy, original thread: http://www.openwall.com/lists/oss-security/2015/09/06/2

phillmv:
Solid. Thanks Justin!

PR #204, opened 2015-09-22
Add CVE-2014-9489 for gollum-grit_adapter

reedloden:
https://github.com/gollum/gollum/issues/913
https://github.com/gollum/grit_adapter/commit/4520d973c81fecfebbeacd2ef2f1849d763951c7


VanessaHenderson:
This also affects gollum-lib and gollum, should we add it for those as well?

PR #207, opened 2015-10-21
heap overflow and information leak in zipruby

grosser:
@phillmv 

https://packetstormsecurity.com/files/111242/libzip-0.10-Heap-Overflow-Information-Leak.html

### DIFF
--- a/gems/zipruby/CVE-2012-1163.yml
+++ b/gems/zipruby/CVE-2012-1163.yml
@@ -1,0 +1,11 @@
+---
+gem: zipruby
+cve: 2012-1163
+url: https://packetstormsecurity.com/files/cve/CVE-2012-1163
+title: A flaw in zipruby may allow information leaks
+date: 2012-03-21
+description: |
+  libzip (version <= 0.10) has two vulnerabilities that may lead to a heap
+  overflow or an information leak via corrupted zip files. PHP (versions
+  5.4.0 and <= 5.3.10) and the Ruby binding zipruby (version <= 0.3.6) are
+  also affected as they include copies of affected libzip versions.
--- a/gems/zipruby/CVE-2012-1162.yml
+++ b/gems/zipruby/CVE-2012-1162.yml
@@ -1,0 +1,11 @@
+---
+gem: zipruby
+cve: 2012-1162
+url: https://packetstormsecurity.com/files/cve/CVE-2012-1163
+title: zipruby is susceptible to a heap overflow
+date: 2012-03-21
+description: |
+  libzip (version <= 0.10) has two vulnerabilities that may lead to a heap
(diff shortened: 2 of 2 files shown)


gems/zipruby/CVE-2012-1163.yml
```diff
@@ -0,0 +1,15 @@
+---
+gem: zipruby
+cve: CVE-2012-1163
```
reedloden: Just "2012-1163" (without the CVE- part)

grosser: done!

gems/zipruby/CVE-2012-1163.yml
```diff
@@ -0,0 +1,15 @@
...
+  libzip (version <= 0.10) has two vulnerabilities that may lead to a heap
+  overflow or an information leak via corrupted zip files
+date: 2012-03-21
+description: |
+  libzip (version <= 0.10) has two vulnerabilities that may lead to a heap
+  overflow or an information leak via corrupted zip files. PHP (versions
+  5.4.0 and <= 5.3.10) and the Ruby binding zipruby (version <= 0.3.6) are
+  also affected as they include copies of affected libzip versions.
+patched_versions:
+  - ">= 0.3.6"
```
VanessaHenderson: There are no versions currently released that are above this, should we really be putting this as patched versions? We can't guarantee that there will ever be a fixed version released...

VanessaHenderson: By that I mean the latest commit in the repository is 2 days before the official release notice of libzip 0.10. Did they actually update it? I don't know much about C...
screen shot 2015-10-22 at 7 41 56 pm
screen shot 2015-10-22 at 7 42 02 pm

sergey-alekseev: should be `> 0.3.6`

phillmv:
OKAY. Thanks for all the pull requests @grosser!

If I understand this correctly, forever ago someone found a couple CVEs in libzip, which is used by zipruby, which has been totally abandoned.

In this light, there aren't any patched versions since https://rubygems.org/gems/zipruby/ shows that 0.3.6 is the last version that has been released, and the advisory claims that 0.3.6 is vulnerable. (Thanks, @VanessaHenderson)

@grosser, if you could please: 
- delete the 'patched_versions' line, since they none exist
- create one file per CVE, with the appropriate description (keep the url the same :))
- while you're at it, change the title of the first one (2012-1162) to "zipruby is susceptible to a heap overflow", and the title of the other (2012-1163) to "A flaw in zipruby may allow information leaks"

and we should be good :).

grosser:
updated, looking good ?

PR #208, opened 2015-10-21
add sidekiq memory leak advisory

grosser:
... not sure how to integrate this ... tried to keep the name unique ... is that good enough ?

@phillmv

### DIFF
--- a/gems/sidekiq/SIDEKIQ-PR-2598.yml
+++ b/gems/sidekiq/SIDEKIQ-PR-2598.yml
@@ -1,0 +1,8 @@
+---
+gem: sidekiq
+url: https://github.com/mperham/sidekiq/pull/2598
+title: Sidekiq Gem for Ruby Leaks memory
+date: 2015-10-09
+description: Sidekiq Gem for Ruby Leaks memory
+patched_versions:
+  - ">= 3.5.1"


reedloden:
You need to get an OSVDB or CVE assigned for this issue. See the directions at http://guides.rubygems.org/security/#reporting-security-vulnerabilities

grosser:
how likely do you think that is ... since it's no a security issue ... ?

reedloden:
I haven't looked into the details, but is it "potential DoS via memory exhaustion"? Really depends how the problem can be caused. Could a user cause problems due to the issue?

reedloden:
But in general, DoS problems of a minor or self-inflicting nature aren't normally classified as security vulnerabilities.

tarcieri:
I would agree that I would class this as a bug and not as a vuln

grosser:
yeah ... sounds about right ... I need a 'these ruby gems have bugs and you should upgrade' database :(

I'll close them out and search for a better way ...

postmodern:
Technically, a memory leak is a potential DoS. If an attacker can degrade or interrupt service, that is a valid attack (ex: memory leak in a software license key server). I would still ask for an OSVDB/CVE ID and see what the response is.

tarcieri:
> Technically, a memory leak is a potential DoS.

What if it happens irrespective of any attacker influence?

No attacker needed == not a vuln in my book

postmodern:
If it aides the attacker or the attacker can influence it (ex: by sending more requests), then it's a vulnerability, imho. I will defer to MITRE/OSVDB though.

tarcieri:
I definitely appreciate the validity of DoS attacks which occur *only* under attacker influence, but if normal system operation causes a self-DoS and all the attacker can do is make it happen marginally faster, I think it's silly to class it as a vuln. It makes about as much sense as classing bugs which allow remote code execution as vulns only if you have remote code execution in the first place.

If you do want to class this as a vuln, I would kindly ask you be very very careful about identifying what versions are actually affected. `">= 3.5.1"` is wrong.

My reaction when our internal tooling autofiles vulns for every single internal application which uses Sidekiq will be to immediately close them, FWIW. This is not the sort of issue we normally track as VULNs internally, and I expect others might not be happy with the "noise" it creates in terms of vulnerability reporting. We have different SLAs for fixing vulns versus bugs, for example.

kurtseifried:
How much faster are we talking? How long does it take an attacker to trigger this issue? Minutes? Hours? Days? vs. how long it takes the issue to naturally surface?

grosser:
From the looks of it it's not a significant leak or it would have been found sooner ...

kurtseifried:
Also the term "memory leak" is overloaded (are we talking memory not being released? The contents of memory being exposed to an attacker? I assume in this case you mean CWE-401

https://cwe.mitre.org/data/definitions/401.html

Improper Release of Memory Before Removing Last Reference ('Memory Leak')?

kurtseifried:
@grosser right but if an attacker can trigger the conditions and bring the system down quickly then it becomes a security problem.

tarcieri:
@kurtseifried I believe this leak in particular is triggered by jobs that crash, i.e. an unhandled exception is raised within a job. So an attacker needs to:

1) find a specific endpoint on a web frontend which puts jobs into Redis, and hit that repeatedly
2) specifically find a way to cause the job type a web frontend is submitting to crash

If I understand the problem correctly, this problem does not manifest under normal system operation, nor is there a generic way for an attacker to trigger it.

As someone who triages a vuln queue professionally, this falls into the "not a vuln - ignore" category for me.

kurtseifried:
So this could definitely get a CVE for the intersection of these vulnerabilities, I would suggest posting on oss-security@ to get a final ruling from Mitre, my gut says give it a CVE because often times these things are more exploitable than we think at first.

phillmv:
In short, @grosser - sounds like you need to be more specific regarding the version of sidekiq being supplied and obtain a CVE. Do let us know if you need help navigating the CVE step.

Thanks for your effort!

PR #210, opened 2015-10-21
grape leaks memory

grosser:
... and more 

@phillmv


postmodern:
Just needs a CVE/OSVDB id.

PR #211, opened 2015-10-21
celluloid

grosser:
@phillmv last one!


tarcieri:
You're classing memory leaks as VULNs? That seems a bit onerous, and also the version range you specified does not accurately reflect which versions are actually affected.

For those of us who have vulnerability monitoring systems which source this data, this will set off vulnerability alarm bells for something that isn't strictly a vuln, and for versions that are unaffected. It will also impact every single Sidekiq user.

postmodern:
I'm curious what MITRE/OSVDB's stance on memory leaks are. /cc @kurtseifried

PR #212, opened 2015-10-26
Fixing OSVDB-90587's patched versions

jeffreyc:
The current advisory uses `~>` to explicitly flag only two versions as patched. This causes automated tools to flag Ruby 2.1.0+ as vulnerable.

[The advisory](http://www.osvdb.org/show/osvdb/90587) states "Upgrade to version 1.9.3-p392, or higher, to address this vulnerability." Ruby 2.0.0.0 was released after 1.9.3-p392, so a simple `>=` is adequate to cover all newer versions.

### DIFF
--- a/rubies/ruby/OSVDB-90587.yml
+++ b/rubies/ruby/OSVDB-90587.yml
@@ -13,4 +13,4 @@ description: |
 cvss_v2: 5.0
 patched_versions:
   - ~> 1.9.3.392
-  - ~> 2.0.0.0
+  - ">= 2.0.0.0"


phillmv:
Not quite - looking into it, https://www.ruby-lang.org/en/news/2013/02/22/json-dos-cve-2013-0269/ states,

>All ruby 2.0 versions prior to ruby 2.0.0 patchlevel 0

I don't think there are many people running Ruby 2 pre versions but for the sake of completeness, instead of deleting 2.0.0.0 want to make it `>=` ?

Out of curiosity, what automated tooling are you using? I didn't think bundler-audit supported this part of the dataset yet.

jeffreyc:
> I don't think there are many people running Ruby 2 pre versions but for the sake of completeness, instead of deleting 2.0.0.0 want to make it >= ?

Sorry, between the referenced OSVDB issue and the Ruby 2.0.0 release timeline, I assumed all 2.0.0 releases were safe. Excellent point, fixed.

> Out of curiosity, what automated tooling are you using? I didn't think bundler-audit supported this part of the dataset yet.

It doesn't, but it's a great candidate to check both Ruby and RubyGems, so I'm working on a PR to add that behavior. I discovered this issue during my testing of the Ruby functionality.

phillmv:
Cool. I'm sure postmodern will be receptive - just be forewarned that there's some inconsistency with how minor and patch versions are reported by the interpreter. 

I'd link you to the issue where it was discussed but I can't seem to find it :P.

reedloden:
https://github.com/rubysec/ruby-advisory-db/pull/45 is what you're referring to, @phillmv

jeffreyc:
Thanks for the pointer. Fortunately, I think you addressed a lot of the trickiness with the decision to use `<major>.<minor>.<bugfix>.preview.<patchlevel>`.

PR #213, opened 2015-11-24
Add advisories for libxml2 and libxslt vulnerabilties patched in Nokogiri

koenrh:
Adds advisories for vulnerabilities in libxml2 and libxslt libraries that have recently been patched in Nokogiri.

Mentioned in release notes for Nokogiri v1.6.7.rc4: https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.rdoc#167rc4--2015-11-22

### DIFF
--- a/gems/nokogiri/CVE-2015-1819.yml
+++ b/gems/nokogiri/CVE-2015-1819.yml
@@ -39,7 +39,8 @@ description: |
   uninitialized value(s)" and unsafe memory access. This issue does not have a
   CVE assigned yet. See related URLs for details. Patched in v1.6.7.rc4.
 patched_versions:
-  - ">= 1.6.6.4"
+  - "~> 1.6.6.4"
+  - ">= 1.6.7.rc4"
 related:
   cve:
     - 2015-7941
--- a/gems/nokogiri/CVE-2015-1819.yml
+++ b/gems/nokogiri/CVE-2015-1819.yml
@@ -1,0 +1,43 @@
+---
+gem: nokogiri
+cve: 2015-1819
+url: https://github.com/sparklemotion/nokogiri/issues/1374
+title: Nokogiri gem contains several vulnerabilities in libxml2 and libxslt
+date: 2015-04-14
+description: |
+  Several vulnerabilities were discovered in the libxml2 and libxslt libraries
+  that the Nokogiri gem depends on.
+
+  CVE-2015-1819
+  A denial of service flaw was found in the way libxml2 parsed XML
+  documents. This flaw could cause an application that uses libxml2 to use an
+  excessive amount of memory.
+
+  CVE-2015-7941
+  libxml2 does not properly stop parsing invalid input, which allows
+  context-dependent attackers to cause a denial of service (out-of-bounds read
+  and libxml2 crash) via crafted specially XML data.
+
+  CVE-2015-7942
+  The xmlParseConditionalSections function in parser.c in libxml2
+  does not properly skip intermediary entities when it stops parsing invalid
+  input, which allows context-dependent attackers to cause a denial of service
+  (out-of-bounds read and crash) via crafted XML data.
+
+  CVE-2015-7995
+  The xsltStylePreCompute function in preproc.c in libxslt 1.1.28 does not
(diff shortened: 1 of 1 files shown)


phillmv:
Fabulous. Suggestion:

Rather than create four different advisory files for a single release in nokogiri, condense them into a single advisory. For the description, use:

```txt
Several vulnerabilities were discovered in the libxml2 and libxslt 
libraries that the Nokogiri gem depends on. 

CVE-2015-1819
A denial of service flaw was found in the way libxml2 parsed XML
documents. This flaw could cause an application that uses libxml2 to use an
excessive amount of memory.

CVE-2015-7941
libxml2 etc etc
```

Pick the first CVE to lead as the CVE key and drop all the other CVEs into a `related` field like so:
```yaml 
related:
  cve: 
    - 2015-1819
    - etc
```

([See here for an example](https://github.com/rubysec/ruby-advisory-db/blob/master/gems/sidekiq/OSVDB-125676.yml))

Barring any suggestions from other people, I'm also OK with keeping the date to the earliest CVE being patched.

Thanks, Koen!

skorth:
@koenrh Take a look at Version 1.6.6.4 and 1.6.6.3.
Take a look at:
https://github.com/sparklemotion/nokogiri/commit/f5b8e91bcabd34d190c3a183ab86398edd75b9b4
and
https://github.com/sparklemotion/nokogiri/commit/2896e941252c611d112a4a11895a7da369b40c10

gems/nokogiri/CVE-2015-1819.yml
```diff
@@ -0,0 +1,43 @@
...
+  The xsltStylePreCompute function in preproc.c in libxslt 1.1.28 does not
+  check whether the parent node is an element, which allows attackers to cause
+  a denial of service using a specially crafted XML document.
+
+  CVE-2015-8035
+  The xz_decomp function in xzlib.c in libxml2 2.9.1 does not
+  properly detect compression errors, which allows context-dependent attackers
+  to cause a denial of service (process hang) via crafted XML data.
+patched_versions:
+  - ">= 1.6.7.rc4"
```
igrep: According to https://github.com/sparklemotion/nokogiri/blob/v1.6.6.4/CHANGELOG.rdoc#1663--2015-11-16 1.6.6.4 is also patched.

koenrh:
@phillmv: Fixed. It makes perfect sense to aggregate these advisories into a single file. 

@skorth: Thanks. You're right, specified CVEs have been fixed in v1.6.6.3 and v1.6.6.3. However, v1.6.7.rc4 also fixes another out-of-bounds memory access issue ( https://github.com/sparklemotion/nokogiri/commit/8f3de6d88d0da11fb62a45daa61b85ce71b4af59) in libxml2. The problem is this particular issue doesn't have a CVE assigned (yet). How do you deal with these kind of issues?

phillmv:
Eh, OK, why don't we bump the patched version 1.6.6.4, make a note of it in the description, and add the other urls to the related field, i.e.

```yaml
related:
  url:
    - relevant-issue
    - relevant-commit-or-etc
```

We'll add in the other CVE when it gets assigned.

phillmv:
Normally I'd be up for making a separate advisory, but since these were all so close together and 1.6.6.3 and 1.6.6.4 are just security updates…

koenrh:
Ok. Set the patched version to v1.6.6.4, and added some details about that out-of-bounds memory access issue that doesn't have a CVE assigned yet.

phillmv:
Sorry Koen, I took a closer peak at how Nokogiri [does versions](https://github.com/sparklemotion/nokogiri/releases). Since there is a `1.6.7.rc`*`3`* we should be more specific in the patched versions.

If you could change it to:
```yaml
patched_versions:
  - "~> 1.6.6.4"
  - ">= 1.6.7.rc4"
```
we should be able to call it a day :).

koenrh:
Ha. Thanks for your help, @phillmv.

Do you want me to squash all commits into one unit?

phillmv:
… maybe in the future but for now I'm just going to merge it in :).

phillmv:
Thanks @igrep @skorth and @koenrh!

PR #217, opened 2015-12-16
Add OSVDB-131677 for mail ruby gem

### LINKED ISSUE #215: Add advisory for SMTP injection vulnerability in mail <2.6.0
Need an advisory for the mail vulnerability discussed in http://www.mbsd.jp/Whitepaper/smtpi.pdf (section 3.1)

Affects <2.6.0

reedloden:
Fixes #215.

### DIFF
--- a/gems/mail/OSVDB-131677.yml
+++ b/gems/mail/OSVDB-131677.yml
@@ -1,0 +1,19 @@
+---
+gem: mail
+osvdb: 131677
+url: http://www.mbsd.jp/Whitepaper/smtpi.pdf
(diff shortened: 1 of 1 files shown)


phillmv:
Suggested rewording:

`Mail rubygem vulnerable to SMTP Injection via recipient email addresses`


Because the Mail rubygem does not validate or impose a length limit on email address fields, an attacker can modify messages sent with the gem via a specially crafted recipient email address.

Applications that validate email address format are not affected by this vulnerability.
  
The recipient attack is described in Terada, Takeshi. "SMTP Injection via Recipient Email Addresses." 2015. The attacks described in the paper (Terada, p. 4) can be applied to the library without any modification.

PR #225, opened 2016-01-07
Add OSVDB-132234 for rack-attack

reedloden:
https://github.com/kickstarter/rack-attack/releases/tag/v4.3.1
https://github.com/kickstarter/rack-attack/commit/76c2e3143099d938883ae5654527b47e9e6a8977

### DIFF
--- a/gems/rack-attack/OSVDB-132234.yml
+++ b/gems/rack-attack/OSVDB-132234.yml
@@ -1,0 +1,26 @@
+---
+gem: rack-attack
+osvdb: 132234
+url: https://github.com/kickstarter/rack-attack/releases/tag/v4.3.1
+title: |
+  rack-attack Gem for Ruby missing normalization before request path
+  processing
+date: 2015-12-18
+description: |
+  When using rack-attack with a rails app, developers expect the request
+  path to be normalized. In particular, trailing slashes are stripped so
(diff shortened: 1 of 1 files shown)


gems/rack-attack/OSVDB-132234.yml
```diff
@@ -0,0 +1,27 @@
+---
+gem: rack-attack
+framework: rails
```
reedloden: Is my understanding of the `framework` field correct here? Using it to mean only when used with Rails are you affected by this issue...

JuanitoFatas: Shouldn't have :blush: ? From README.md:

> framework [String] (optional): Name of framework gem belongs to.
> [Schema](https://github.com/rubysec/ruby-advisory-db#schema)

reedloden: Good catch. We should fix `twitter-bootstrap-rails` then.

reedloden:
@JuanitoFatas / @VanessaHenderson, any other comments, or looks good? :)

VanessaHenderson:
Other than the platform comment it LGTM to me! ;)

skorth:
Looks good :+1:

reedloden:
Thanks!

PR #232, opened 2016-01-18
Add advisories for git-fastclone

reedloden:
WIP -- missing OSVDB/CVE entry for one advisory

### DIFF
--- a/gems/git-fastclone/OSVDB-133534.yml
+++ b/gems/git-fastclone/OSVDB-133534.yml
@@ -1,0 +1,13 @@
+---
+gem: git-fastclone
+osvdb: 133534
+url: https://github.com/square/git-fastclone/pull/5
+title: git-fastclone Shell Metacharacter Injection Arbitrary Command Execution
+date: 2015-12-15
+description: |
+  git-fastclone before 1.0.5 passes user modifiable strings directly to a shell
+  command. An attacker can execute malicious commands by modifying the strings
+  that are passed as arguments to "cd " and "git clone " commands in the
+  library.
+patched_versions:
+  - ">= 1.0.5"
--- a/gems/git-fastclone/OSVDB-133535.yml
+++ b/gems/git-fastclone/OSVDB-133535.yml
@@ -1,0 +1,21 @@
+---
+gem: git-fastclone
+osvdb: 133535
+url: https://github.com/square/git-fastclone/pull/2
+title: git-fastclone permits arbitrary shell command execution from .gitmodules
+date: 2015-12-11
+description: |
+  Git allows executing arbitrary shell commands using git-remote-ext via a
+  remote URLs. Normally git never requests URLs that the user doesn't
+  specifically request, so this is not a serious security concern. However,
+  submodules did allow the remote repository to specify what URL to clone
+  from.
+
+  If an attacker can instruct a user to run a recursive clone from a
+  repository they control, they can get a client to run an arbitrary shell
+  command. Alternately, if an attacker can MITM an unencrypted git clone,
+  they could exploit this. The ext command will be run if the repository is
+  recursively cloned or if submodules are updated. This attack works when
+  cloning both local and remote repositories.
+patched_versions:
+  - ">= 1.0.1"


gems/git-fastclone/OSVDB-1.yml
```diff
@@ -0,0 +1,11 @@
+---
+gem: git-fastclone
+osvdb: 
+url: https://github.com/square/git-fastclone/pull/5
+title: |
+  git-fastclone Improper use of cocaine allows for malicious common injection
+date: 2015-12-15
+description: |
+  Fastclone is using cocaine improperly, allowing malicious command injection.
```
reedloden: need a better description

reedloden:
@bburky / @longboardcat / @alokmenghrajani -- a few questions

Has a CVE/OSVDB ID been requested for https://github.com/square/git-fastclone/pull/5 yet?

Are the reports submitted to HackerOne going to be publicly disclosed? Would be useful for referencing.

I haven't seen any advisories about these issues. Just making sure I'm not missing something.

bburky:
The other issue disclosed on hackerone was fixed in square/git-fastclone#2 and square/git-fastclone#3. It was an independent issue, but the same problem that was fixed in git core in CVE-2015-7545.

@longboardcat, @alokmenghrajani I'd be fine with making the HackerOne issues public now that they have been resolved.

bburky:
To my knowledge a CVE has not been requested for either of these issues fixed in git-fastclone.

longboardcat:
That's okay with me. What do I need to do?

James

reedloden:
@longboardcat -- referring to HackerOne or requesting CVE/OSVDB?

For the former, in the report, select "Publicly disclose" or "Request public disclosure" or whatever the right option is from the drop-down box near the comment box. @bburky will need to do the same on his end so the reports are disclosed.

For the latter, follow the directions outlined at http://guides.rubygems.org/security/#reporting-security-vulnerabilities -- specifically, I would e-mail oss-security@ and cve-assign@ with the information, requesting a CVE. I would then forward that request to OSVDB to get an assignment there as well. I'm happy to help you with this if you would like.

gems/git-fastclone/OSVDB-1.yml
VanessaHenderson: I looked at the GitHub Fix Commit and came up with this description https://srcclr.com/security/command-injection/ruby/s-1922

reedloden:
I submitted http://seclists.org/oss-sec/2016/q1/166 for getting CVEs assigned. Also notified OSVDB.

longboardcat:
Thanks! Haven't gotten around to any of this :(

James

VanessaHenderson:
Hmmm, the only thing I can see if that git-fastclone hasn't actually pushed any packages to RubyGems since September 2015, where as both of these pulls were merged Dec 2015.
I just downloaded the latest version that was pushed to RubyGems 1.0.14 and it doesn't contain either fix. Someone should nag them to push to RubyGems (and probably with a more accurate version number)

reedloden:
@VanessaHenderson great catch. I made sure the version was uploaded, but I didn't notice the date. My bad.

@longboardcat what's up with the rubygems mess? Can it be cleaned up to match reality?

longboardcat:
The latest version should be correct, there were some mistakes earlier that
were pulled from the gem site. I actually kept the date the same in the
gemspec, but I can push a new version with the new date.

James

VanessaHenderson:
@longboardcat what about the fix for the OSVDB-133535 the pull indicated adds a log message and an if statement to the code https://github.com/square/git-fastclone/blob/14198fe12443055839b1ba4cc294b04a38ae15f1/lib/git-fastclone.rb#L250 which is not present anymore in master https://github.com/square/git-fastclone/blob/master/lib/git-fastclone.rb#L296 

Looking at the code it seems the implementation has changed (where my confusion was caused) https://github.com/square/git-fastclone/commit/93a4634abca94464387a8bde5a3062e858fc0f1e#diff-366e68110b1de1ebeb9dc33605fd697fL249 so perhaps this should be the fix commit listed instead of the  PR since those changes were quickly overwritten. 

Same with the other OSVDB link and https://github.com/square/git-fastclone/commit/ac3dd9871ff1722b6b4f3c6c7a1250b81a08644f

I feel like these commits should be link because if people are using these commits for patches, they are going to use the not-current fixes for the patches, if that makes sense

bburky:
Yes, square/git-fastclone@93a4634 replaces square/git-fastclone@14198fe with a better fix using the GIT_ALLOW_PROTOCOL environment variable newly introduced into git itself. If you can link to both square/git-fastclone#2 and square/git-fastclone#3 that would be best.

@longboardcat Can we make https://hackerone.com/reports/104465 and https://hackerone.com/reports/105190 public please? Without the POCs it isn't terribly obvious that this is remotely exploitable with an attacker controlled git repo or git server.

longboardcat:
I can't even see them for some reason. I'll ask ping the Square employee
who might help tomorrow.

James

aviat:
Interesting : people complaining about a non public hacker one Ruby gem vuln reported to square.

reedloden:
@aviat what's the issue? The vulnerabilities were reported to Square via https://hackerone.com/square-open-source, of which the `git-fastclone` ruby gem is in scope.

longboardcat:
The reports should be public now.

James

VanessaHenderson:
I can confirm they are public :+1:

reedloden:
Sorry for the ridiculous delay on these.

HackerOne has assigned CVEs for the two issues here:
CVE-2015-8968 -- https://hackerone.com/reports/104465
CVE-2015-8969 -- https://hackerone.com/reports/105190

I'll update the advisories and get this pushed.

mveytsman:
@reedloden it looks like the reports are up: https://hackerone.com/reports/104465 and https://hackerone.com/reports/105190

Do you mind updating the advisories with the CVE's and links to the reports?

*sigh* I wish github made it easier for me to just push to your PR :(

mveytsman:
Moved to #277

PR #235, opened 2016-01-25
Batch Rails and related vulns from 2016-01-26

phillmv:
Review welcome.

PR #236, opened 2016-01-26
adds CVE-2015-7578

nealharris:
cc: @tenderlove


nealharris:
Just saw there's already https://github.com/rubysec/ruby-advisory-db/pull/235, which includes more than this.  Sorry for the noise.

PR #240, opened 2016-02-29
Add CVE-2016-2097 for actionview

thirstscolr:


phillmv:
thanks, @thirstscolr!

PR #241, opened 2016-02-29
Add CVE-2016-2098 for actionpack

thirstscolr:


phillmv:
At a glance looks fine - just missing `unaffected_versions` - i.e. see [here](https://github.com/rubysec/ruby-advisory-db/blob/master/gems/actionpack/CVE-2015-7581.yml). 

Throw that in here and in #240 and we should be good to go!

thirstscolr:
Added `unaffected_versions` here and #240

thirstscolr:
@phillmv if everything LGTY can we merge this and #240?

PR #245, opened 2016-03-09
Use optimistic version constraint for Rails-related CVEs

dcarral:
Fix issues #244 & #140 following the approach @reedloden applied @ https://github.com/rubysec/ruby-advisory-db/commit/782f008ee01c25bff1bea2c357de5162a3132cd5 to fix #243.

The idea is to whitelist the recently released v4.2.6 Rails version.

PR #247, opened 2016-03-09
Unlock explicit version of RSpec and add Gemfile.lock

### LINKED ISSUE #246: Incompatibility between RSpec 3.3.2 & Rake 11.0
Quoting @jekuta's comment @ PR #245:

> The Travis build failing because there is an issue between RSpec 3.3.2 and Rake 11.0, which don't seem to be compatible.
> 
> RSpec::Core:::RakeTask uses a Rake.application.last_comment method that was removed from Rake 11. It is still on Rake 10.5.0.
> 
> https://github.com/rspec/rspec-core/blob/v3.3.2/lib/rspec/core/rake_task.rb#L82
> https://github.com/ruby/rake/blob/v10.5.0/lib/rake/task_manager.rb

dcarral:
This fixes #246, which is the reason why the Travis build for #245 is failing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
-Gemfile.lock
 _site
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rspec', '3.3.0'
+gem 'rspec'
 gem 'rake'
 
 group :development do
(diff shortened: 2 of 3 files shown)


reedloden:
Thanks!

dcarral:
Thank you for merging this! Looking forward to mute the `bundler-audit` warnings when upgrading to Rails v4.2.6 :)

PR #248, opened 2016-03-09
Allow Rails 4.1.15+ for 4.1.14.1 and 4.1.14.2 CVEs

matt-glover:
Related to https://github.com/rubysec/ruby-advisory-db/pull/245 but this is for the 4.1 line.

The old version constraint considered version 4.1.15 vulnerable to security issues patched in 4.1.14.1 or 4.1.14.2. Changed it to treat everything newer than 4.1.14 in the 4.1 line as patched.


gems/actionpack/CVE-2015-7576.yml
```diff
@@ -112,5 +112,5 @@ description: |
 patched_versions:
   - "~> 5.0.0.beta1.1"
   - "~> 4.2.5.1" 
-  - "~> 4.1.14.1"
+  - "~> 4.1.14, >= 4.1.14.1"
```
matt-glover: I am assuming this is the correct syntax for a two-part constraint based on previous use [here](https://github.com/rubysec/ruby-advisory-db/blob/de575b873334e0a1eaab6d9a8dca3ea059bbc7a4/gems/actionpack/OSVDB-84243.yml#L22). A test run of `bundler-audit` seemed happy with the syntax.

postmodern: @matt-glover yep! bundler-audit [splits by `, `](https://github.com/rubysec/bundler-audit/blob/0961853eac0e64da45b4276b0a6408469964d099/lib/bundler/audit/advisory.rb#L54).

Issue #224, opened 2016-01-05
RubySec numbering system


mveytsman:
We should stop relying on osvdb and cve ids as canonical identifiers as they may not be always present, as in the case where we get a vuln before it's in either db.
- [ ]  Add a new field to the schema, `rubysec-id`
- [ ] Come up with numbering scheme. My suggestion is to model after CVE, ala `RUBYSEC-2016-00001`
- [ ] Add a `rubysec-id`s to past vulns in the database.
- [ ] Come up with policy for generating new id numbes are vulns are added. This should be done by the person with commit bit who's accepting the PR.

reedloden:
How do we track what is the next available ID to use?

For the year assignment, is that the date of the disclosure or date of finding? Depending on that, need to be able to continually add new IDs for previous years as vulnerabilities are found (that we weren't tracking).

Proper CVE-like format would be `RUBYSEC-YYYY-NNNN` where NNNN is always at least 4 numbers, zero-padded (can be increased beyond that over time).

Does `rubysec-id` make sense when we use `cve` and `osvdb` right now? Perhaps just `rubysec` for the field? or `rsid` or something...

tarcieri:
> How do we track what is the next available ID to use?

Each time IDs are assigned, you could check them into a registry file in the repo (perhaps with a note about whom they were assigned to)

phillmv:
>How do we track what is the next available ID to use?

I think we should script something easy to accept PRs with - maybe it'll calc the new id, merge the changes, apply new id then push the merged results, rather than just using GH.

>For the year assignment, is that the date of the disclosure or date of finding?

I vote disclosure. Keep it simple. 

>Does rubysec-id make sense when we use cve and osvdb right now? Perhaps just rubysec for the field? or rsid or something...

I'd propose RBS-YYYY-NNNN and keep it `rubysec` or `rubysec-id`. 

In my heart of hearts, I think it'd be cool to ditch the centrally managed monotonic integer and instead have some kind of hashing scheme to keep id-space management simple.

Suppose we have YYYY- {7 most significant bits of sha1(package, date) (increment by one in case of collision)}. I can't be arsed to calc the probability of collision at our volume but I'm going to guess it's rather low. We still have to write the script to check on merge, but by and large anyone can generate an ID and it becomes valid by being accepted into the repo. (I totally get it if people think that's too much of a pain :P)

JonRowe:
Why not write a script to calculate it and apply as a build step via travis-ci?

[...]

PR #252, opened 2016-04-08
First pass at a rubysec id reservation scheme

mveytsman:
`rake reserve_id` reserves an id

`rake reserve_id[2012]` reserves an id in 2012

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,24 @@ the advisories [OSVDB] identifier number.
         OSVDB-79727.yml  OSVDB-84513.yml  OSVDB-89026.yml  OSVDB-91454.yml
         OSVDB-84243.yml  OSVDB-84515.yml  OSVDB-91452.yml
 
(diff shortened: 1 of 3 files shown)


postmodern:
How do we avoid collisions where two contributors reserve the same ID?

mveytsman:
My assumption was that you do the reservation in an earlier commit/PR and handle merge conflicts by hand. 

Can you think of a better way?

postmodern:
Also do we really need a CVE-like ID for advisories? Advisories are currently grouped by effected gem or ruby, so the IDs would only need to be unique in that scope.

mveytsman:
My worry about scoping to gem is the edge case when gems change names.

postmodern:
Ah, well the previously released versions of those gems would still have the same name. So we'd just make a new directory for the new gem name.

mveytsman:
Closing this see #224

PR #253, opened 2016-04-24
Add safemode ruby gem advisory

skorth:
Please double check, thx.

### DIFF
--- a/gems/safemode/CVE-2016-3693.yml
+++ b/gems/safemode/CVE-2016-3693.yml
@@ -1,0 +1,14 @@
+---
+gem: safemode
+cve: 2016-3693
+title: Information leakage through templates
+date: 2016-04-20
+url: http://seclists.org/oss-sec/2016/q2/119
+description: |
+  A provisioning template containing `inspect` will expose sensitive
+  information about the Rails controller and application when using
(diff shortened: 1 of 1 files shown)


gems/safemode/CVE-2016-3693.yml
```diff
@@ -0,0 +1,14 @@
+---
+gem: safemode
+cve: 2016-3693
+title: Information leakage through templates
+date: 2016-04-20
+url: http://seclists.org/oss-sec/2016/q2/119
+description: |
+  A provisioning template containing `inspect` will expose sensitive
```
reedloden: This is all Foreman-specific, but the gem has nothing to do with foreman. Can you rewrite this to be specific to what was actually vulnerable for `safemode`?

skorth: @reedloden suggestion:

"Safemode is initialised with an optional 'delegate' object.
If the delegated object is a Rails controller, 'inspect' could
be called which then exposes all informations about the App,
including routes, secret tokens, caches and so on."

reedloden: Seems fine. Update the title as well?

skorth:
@reedloden Sorry for the delay -> updated the title.

PR #256, opened 2016-05-07
Add advisory for espeak-ruby

bcoles:
- Ancient bug
- ~~No response from vendor~~ Now patched in version 1.0.3
- No CVE


phillmv:
Thanks!

The way this db is set up we can't accept it without a cve - but this PR prompted action on another ticket to get around that. Once we resolve it in the next few days, we'll circle back to here.

Cheersm

mveytsman:
I have requested a CVE for this

mveytsman:
Moved to #276

PR #258, opened 2016-05-28
Update CVSSv2 link in README

riffraff:
The current link is dead, the proposed one might not have exactly the same content, but it seems at least related.


reedloden:
Thanks!

PR #260, opened 2016-06-17
Ruby >= 2.3.0 has been patched for CVE-2015-1855

mattbrictson:
CVE-2015-1855 was patched for Ruby 2.2.2 and all future versions of Ruby. That means in addition to `~> 2.2.2` being patched, `>= 2.3.0` is patched as well.

### DIFF
--- a/rubies/ruby/CVE-2015-1855.yml
+++ b/rubies/ruby/CVE-2015-1855.yml
@@ -14,4 +14,4 @@ description: |
 patched_versions:
   - ~> 2.0.0.645
   - ~> 2.1.6
-  - ~> 2.2.2
+  - ">= 2.2.2"


rubies/ruby/CVE-2015-1855.yml
```diff
@@ -15,3 +15,4 @@ patched_versions:
   - ~> 2.0.0.645
   - ~> 2.1.6
   - ~> 2.2.2
+  - ">= 2.3.0"
```
reedloden: Any reason why not to just change the `~> 2.2.2` to `">= 2.2.2"`?

mattbrictson: I was just following the convention of this commit: 51c76a755d2955ca1e621dd0c4b9a2de3134d9c5 . But yes, I like `">= 2.2.2"` better. I'll update the PR.

reedloden: Ah, are Ruby 2.3.0 preview 1 or preview 2 vulnerable to this?

mattbrictson:
PR updated.

reedloden:
See my other follow-up question. Sorry for not asking that earlier.

mattbrictson:
The [announcement](https://www.ruby-lang.org/en/news/2015/04/13/ruby-openssl-hostname-matching-vulnerability/) makes no mention any 2.3 preview releases.

> All Ruby 2.0 versions prior to Ruby 2.0.0 patchlevel 645
All Ruby 2.1 versions prior to Ruby 2.1.6
All Ruby 2.2 versions prior to Ruby 2.2.2
prior to trunk revision 50292

2.3.0-preview1 was published several months after this announcement, so it is safe to assume it has been patched. The announcement was in April 2015; 2.3.0-preview1 was released in November 2015.

So I think `>= 2.2.2` is adequate.

reedloden:
Awesome. Thanks for checking!

PR #266, opened 2016-08-13
Fix missing actionview vulnerabilities for Rails 3.2 and Rails 4.0

GUI:
Since actionview was not a separate gem prior to Rails 4.1, the following CVEs were not being reported by bundler-audit for Rails 3.2 or Rails 4.0:
- CVE-2016-0752
- CVE-2016-2097
- CVE-2016-6316

This resolves the issue by creating copies of those CVEs, but for the actionpack gem, where the actionview code lived in Rails 4.1 and prior.

I'm not exactly sure this duplicate file approach is the best solution, so let me know if you have any other suggestions for how to manage this. But after the Rails security updates yesterday, I was surprised when bundler-audit still didn't report any new issues for some Rails 3.2 apps, so I wanted to try to fix this in some fashion.

### DIFF
--- a/gems/actionpack/CVE-2016-0752.yml
+++ b/gems/actionpack/CVE-2016-0752.yml
@@ -1,5 +1,5 @@
 ---
-gem: actionview
+gem: actionpack
 framework: rails
 cve: 2016-0752
 date: 2016-01-25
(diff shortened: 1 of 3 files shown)


phillmv:
Oh, great find! Thanks for this.

I'm mid travel, will kick the tires on this later this week - there may be issues with duplicate filenames.

phillmv:
We've handled something like this before, tho which examples in particular don't come immediately to mind.

jasnow:
@phillmv  - Can this issue be closed as done?

PR #274, opened 2017-01-31
Directory traversal vulns for minitar

mveytsman:
Thanks @ecneladis!


VanessaHenderson:
Looks GTM from my side!

raymondberg:
I think there's an issue with this merge. The version specified is `0.60` but the version to fix is `0.6`.  Is there a good way to address this? Checks are failing even after an upgrade to `0.6`.

mveytsman:
That is my fault. I will update with the correct patched version.

Thanks @raymondberg

mveytsman:
https://github.com/rubysec/ruby-advisory-db/commit/a30d20526e013654c0ba832a50f7f48d2c1c4338

PR #279, opened 2017-02-21
ActionView and ActiveRecord 4.2.8 are patched for these vulnerabilities

DouweM:
`bundle-audit` was failing like so, because it didn't consider 4.2.8 as having patches for these 2 vulnerabilities:

```sh
Updating ruby-advisory-db ...
Cloning into '/root/.local/share/ruby-advisory-db'...
Updated ruby-advisory-db
ruby-advisory-db: 282 advisories
Name: actionview
Version: 4.2.8
Advisory: CVE-2016-6316
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/I-VWr034ouk
Title: Possible XSS Vulnerability in Action View
Solution: upgrade to ~> 4.2.7.1, >= 5.0.0.1

Name: activerecord
Version: 4.2.8
Advisory: CVE-2016-6317
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/rgO20zYW33s
Title: Unsafe Query Generation Risk in Active Record
Solution: upgrade to ~> 4.2.7.1

Vulnerabilities found!
```

Based on https://github.com/rubysec/ruby-advisory-db/blob/master/gems/activerecord/CVE-2015-7577.yml#L104-L107, I _think_ this patch should do it, and allow actionview 4.2.7.1, 4.2.8 and 5.0.0.1, but not 5.0.0.

### DIFF
--- a/gems/actionview/CVE-2016-6316.yml
+++ b/gems/actionview/CVE-2016-6316.yml
@@ -51,5 +51,6 @@ unaffected_versions:
 
 # "~> 3.2.22.3" is found in gems/actionpack/CVE-2016-6316.yml
 patched_versions:
-  - ~> 4.2.7.1
+  - "~> 4.2.7.1"
+  - "~> 4.2.8"
   - ">= 5.0.0.1"
--- a/gems/activerecord/CVE-2016-6317.yml
+++ b/gems/activerecord/CVE-2016-6317.yml
@@ -70,4 +70,4 @@ unaffected_versions:
   - ">= 5.0.0"
 
 patched_versions:
-  - ~> 4.2.7.1
+  - ">= 4.2.7.1"


DouweM:
Wow, this took me a lot of commits. We should probably squash it before merging.

mveytsman:
I see you did ` "~> 4.2.7.1, >= 4.2.8"` and '>= 4.2.7.1 for the other. 

Is there a semantic difference there or can you just standardize on the first?

DouweM:
@mveytsman I _think_ that with ` 5.0.0.1` and `"~> 4.2.7.1, >= 4.2.8"`, `5.0.0` wouldn't be incorrectly deemed patched.

mveytsman:
Ah I see @DouweM 

The list of requirements is ORed, while two requirements separated by a comma are ANDed

So in this case `"~> 4.2.7.1, >= 4.2.8"` means versions that are `"~> 4,2.7.1"` AND `">= 4.2.8"`, which is nothing.

I believe the correct patched versions are 

```
  - "~> 4.2.7.1"
  - "~> 4.2.8"
 - ">= 5.0.0.1"
```

Can you update to those and squash the commits and I'll merge.

Thanks!

DouweM:
@mveytsman Heh, I really messed that up. Thanks for your help; updated!

PR #283, opened 2017-02-24
Remove `warning: constant ::Fixnum is deprecated`

spickermann:
This removes a deprecation warning introduced by Ruby 2.4.0 that appears when running specs.

This fix works, because `Integer` exists in older versions of Ruby too and is a superclass of `Fixnum`.


mveytsman:
Thanks for fixing the spacing too :)

spickermann:
Haha, I suggest [files?w=1](283/files?w=1)

PR #284, opened 2017-02-24
Rails `5.1.0.alpha` bases on `5.0.0.rc1`

spickermann:
Rails `5.1.0.alpha` was released yesterday and `bundler:audit` reports some issues that were already fixed in `5.0.0.beta1.1`.

Since `5.1.0.alpha` bases on `5.0.0.rc1` (see: https://github.com/rails/rails/commit/8ecc5ab1d88532a239f17c7520ed922c7579b01c) I am pretty sure that we can change the entries in the list of `patched_versions` from `~> 5.0.0.beta1.1` to `>= 5.0.0.beta1.1` to include Rails `5.1.0`.

I suggest [files?w=1](284/files?w=1) to review changed files.

### DIFF
--- a/gems/actionpack/CVE-2015-7576.yml
+++ b/gems/actionpack/CVE-2015-7576.yml
@@ -8,109 +8,109 @@ url: "https://groups.google.com/forum/#!topic/rubyonrails-security/ANv0HDHEC3k"
 title: Timing attack vulnerability in basic authentication in Action Controller.
 
 description: |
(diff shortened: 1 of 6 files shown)


mveytsman:
Thanks @spickermann 

This looks like it can be a problem in the future, and we should always make sure that there's a `>=` constraint (https://github.com/rubysec/ruby-advisory-db/issues/285)

PR #288, opened 2017-04-07
Update OSVDB-109206.yml

mcfiredrill:
fix broken link


phillmv:
Indeed that new link does work. thanks for the pr!

PR #290, opened 2017-05-11
Add entry for nokogiri 1.7.2

chrismo:


phillmv:
@chrismo this is a very well done submission; idk about the other maintainers but this caught me during travel and I'm sorry this took so long to approve :).

The only thing I'd change is the `related` field should capture the canonical urls instead of duplicating the url: main field, but I'll change that myself.

PR #292, opened 2017-06-10
Update patched versions for mail gem vulnerability OSVDB-131677

cjlarose:
See 

https://github.com/mikel/mail/pull/1097

and 

https://hackerone.com/reports/137631

The vulnerability seems to have actually been addressed by versions 2.5.5 and 2.6.6. Please verify that my `patched_versions` specifiers reflect this correctly.

### DIFF
--- a/gems/mail/OSVDB-131677.yml
+++ b/gems/mail/OSVDB-131677.yml
@@ -16,4 +16,4 @@ description: |
   Recipient Email Addresses." 2015. The attacks described in the paper (Terada,
   p. 4) can be applied to the library without any modification.
 patched_versions:
-- ">= 2.6.0"
+- ">= 2.5.5"


gems/mail/OSVDB-131677.yml
```diff
@@ -16,4 +16,5 @@ description: |
   Recipient Email Addresses." 2015. The attacks described in the paper (Terada,
   p. 4) can be applied to the library without any modification.
 patched_versions:
-- ">= 2.6.0"
+- "~> 2.5.5"
+- ">= 2.6.6"
```
jeremy: `>= 2.5.5` is sufficient. 2.6.0-4 are unaffected due to a coincidental side effect of a separate change.

jeremy:
References #215

jeremyolliver:
I don't think `>= 2.5.5` as the only specifier is correct, given that `2.5.5` and `2.6.6` are the patched versions, as it would incorrectly mark 2.6.0 through to 2.6.5 as "fixed", but they don't include the patch. I think this would be more appropriate:

```
patched_versions:
- "~> 2.5.5"
- ">= 2.6.6"
```

jeremy:
@jeremyolliver 2.6.0+ do not include the _explicit_ fix, but they are unaffected due to a coincidental [side effect of a different change](https://github.com/mikel/mail/commit/72befdc4dab3e6e288ce226a7da2aa474cf5be83). So `>= 2.5.5` does cover the versions which are not vulnerable.

phillmv:
Excellent, thank you all very kindly.

PR #293, opened 2017-06-12
CVE-2015-9096 for Ruby SMTP injection. Migrate Mail OSVDB-131677 to CVE-2015-9097.

jeremy:
Add CVE-2015-9096 for Ruby Net::SMTP command injection.

Migrate Mail OSVDB-131677 to CVE-2015-9097.

/cc @unak @shugo @reedloden @phillmv

### DIFF
--- a/gems/mail/OSVDB-131677.yml
+++ b/gems/mail/OSVDB-131677.yml
@@ -1,23 +1,26 @@
 ---
 gem: mail
+cve: 2015-9097
 osvdb: 131677
-url: http://www.mbsd.jp/Whitepaper/smtpi.pdf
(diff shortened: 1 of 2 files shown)


phillmv:
One caveat: moving off the root OSVDB file path will break any links made to https://rubysec.com/advisories/mail-OSVDB-131677, which is listed in a bunch of places, i.e. https://nvd.nist.gov/vuln/detail/CVE-2015-9097

Can we:
1. Keep the old OSVDB file, an add the CVE as a root key in the yaml file
2. Add the new CVE advisory?

phillmv:
Otherwise this is great!

jeremy:
Great point. I'll leave the original OSVDB naming and reference the CVE.

jeremy:
Done.

PR #308, opened 2017-11-17
Add cve-2012-6134 for omniauth-oauth2 gem

tekin:
See https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2012-6134 and also discussion on PR thread to see confirmation that the first patched version is actually 1.1.1, despite what the CVE says:
https://github.com/intridea/omniauth-oauth2/pull/25.

### DIFF
--- a/gems/omniauth-oauth2/CVE-2012-6134.yml
+++ b/gems/omniauth-oauth2/CVE-2012-6134.yml
@@ -1,0 +1,23 @@
+---
+gem: omniauth-oauth2
+cve: 2012-6134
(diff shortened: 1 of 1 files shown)


VanessaHenderson:
Duplicate of [this entry](https://github.com/rubysec/ruby-advisory-db/blob/master/gems/omniauth-oauth2/OSVDB-90264.yml)
I would suggest updating this entry to have a non-OSVDB reference if anything. We don't want to add duplicate entries

tekin:
Ah, my bad, I thought I'd checked but clearly didn't check well enough! I'll close this PR, agree, a duplicate would not be helpful.

PR #318, opened 2018-01-08
Fix the patched version specifier for CVE-2017-8418

aselder:


mveytsman:
Thanks @aselder, I just fixed this in https://github.com/rubysec/ruby-advisory-db/commit/5657381795701707ea74299cd995814aba2bf81a so I'll close this PR.

PR #320, opened 2018-01-11
Removing patched version for Redcloth Gem

greggawatt:
Patched version is wrong as version 4.3.2 still has XSS issues, There is no currently patched version.


greggawatt:
proof: https://github.com/jgarber/redcloth/issues/46

VanessaHenderson:
Under this circumstances in the past wouldn't a new CVE be requested since this CVE is from 2012?

mveytsman:
I don't think this is an issue as stated: https://github.com/jgarber/redcloth/issues/46#issuecomment-361655921

greggawatt:
I agree with @mveytsman's assessment.

PR #336, opened 2018-04-13
Add CVE-2017-18258 for a libxml bug in nokogiri

mveytsman:


phillmv:
Wanna just add this as an other identifier to https://github.com/rubysec/ruby-advisory-db/blob/master/gems/nokogiri/CVE-2017-15412.yml ?

mveytsman:
ya that makes sense

PR #339, opened 2018-06-09
Remove pre-releases from unaffected versions

wktk:
- `2.0.0.beta[1-2]` and `2.0.0.rc[1-6]` are also affected, but accidentally included to unaffected versions as they are `< 2.0.0`.
- In the commit history, `2.0.0-alpha` exists before `2.0.0.beta1` and it's not affected.  Since rubygems treats `2.0.0-alpha` as between `2.0.0.beta2` and `2.0.0.rc1`, we need to explicitly add it to unaffected versions.


reedloden:
Thank you so much!

PR #341, opened 2018-07-05
Remove CVE-2016-10545

ghiculescu:
According to https://github.com/erikhuda/thor/issues/514 nobody thinks this is actually a vulnerability. It seems unlikely to ever get "patched".


JGrishey:
This is the sticking point of **CVE-2016-10545**: "Documentation does not warn that source parameter is vulnerable to malicious input." That issue has now been resolved as of erikhuda/thor#611.

ghiculescu:
I don't think everyone who uses `bundle audit` should be forced to update `thor` to get a new README, so I still think this should be removed. But I'm happy to instead update if there's a new thor release on rubygems?

reedloden:
Agreed, and sorry for the hassles here.

I was going back through unprocessed submissions, and while I did confirm that the issue was still valid in the current `master` branch, the actual potential for this is low based on how this gem is utilized.

I will note that it would have been nice if the `thor` team had not left https://github.com/erikhuda/thor/issues/514 open without any comments for over two years. Makes it difficult to understand how upstream views an issue if things are ignored for so long (makes me err on the side of caution of considering it valid).

PR #343, opened 2018-07-16
Add CVE-2018-1000211 for Doorkeeper

f3ndot:
~~Merge pending the release of 5.0.0.rc2 and 4.4.0: https://github.com/doorkeeper-gem/doorkeeper/pull/1120#issuecomment-405283007~~ they are now released

### DIFF
--- a/gems/doorkeeper/CVE-2018-1000211.yml
+++ b/gems/doorkeeper/CVE-2018-1000211.yml
@@ -1,0 +1,39 @@
+---
+gem: doorkeeper
+cve: 2018-1000211
+date: 2018-07-11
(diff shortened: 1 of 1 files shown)


phillmv:
Dope, thanks. 

@f3ndot how serious would say this vuln is?

f3ndot:
@phillmv I think it's debatable/matter of perspective on the severity of this vuln.

I liken it to the logout button not working on a website, so that's pretty severe in terms of things not working as expected. However, in order to take advantage, the access and/or refresh tokens must be compromised-- an already difficult feat.

Ignoring the ease of which an attacker can get into the appropriate pre-conditions to leverage the attack, it's medium/high because the entire feature is, in fact, non-functional for all public apps.

PR #357, opened 2018-10-17
CVE-2018-16395: OpenSSL::X509::Name equality check does not work 

corymcdonald:
First time filing one of these out, not sure how to target specific ruby versions with the yaml 😕 

This originated here
https://www.ruby-lang.org/en/news/2018/10/17/openssl-x509-name-equality-check-does-not-work-correctly-cve-2018-16395/


reedloden:
Thanks for the submission! Ended up merging #359, as it was more complete and passed tests.

PR #365, opened 2018-10-30
Rename CVE-2018-16468.yaml to CVE-2018-16468.yml

blaze182:
bundler-audit skips `.yaml` files, only `.yml` should be accepted for now

PR #366, opened 2018-10-30
Add specs to check against .yaml files

blaze182:
I suggest adding extra check for filenames to avoid accidental pushing of `.yaml` files, which are ignored in bundler-audit. This results in missed advisories (which is a vulnerability itself 💣, see #365)

Thanks guys for keeping our apps secure 👍


blaze182:
and yesss it is expected to fail now ☯️

blaze182:
Closing in favor of #367

PR #378, opened 2019-03-05
Add new CVEs from rubygems

renatolond:
Disclosed at https://blog.rubygems.org/2019/03/05/security-advisories-2019-03.html

It seems even though they have CVEs at the post, they are not yet in the mitre db.

### DIFF
--- a/libraries/rubygems/CVE-2019-8320.yml
+++ b/libraries/rubygems/CVE-2019-8320.yml
@@ -1,0 +1,20 @@
+---
+library: rubygems
+cve: 2019-8320
+url: https://blog.rubygems.org/2019/03/05/security-advisories-2019-03.html
+title: Delete directory using symlink when decompressing tar
+date: 2019-03-05
+description: |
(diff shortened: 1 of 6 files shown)


maxfelsher:
The blog post announcing the vulnerabilities implies these issues were fixed in 2.7.8 (released in November) but https://blog.rubygems.org/2019/03/05/2.7.9-released.html describes the vulnerabilities as addressed in 2.7.9 (released the same day as the post about the vulnerabilities). I'm assuming that 2.7.9 is actually correct (which would require a change to the YAML files here) but I'm not 100% sure.

renatolond:
It does look like, I think the 2.7.8 was a typo. I'll fix the yml files

reedloden:
Patched versions should be a range, not a specific release... So, "~> x.y.z" or ">= x.y.z" depending on the specific version.

renatolond:
I didn't notice that before, thanks! Fixed it.

reedloden:
Thank you so much!

PR #379, opened 2019-03-13
Add CVE 2019-5419

greysteil:
That.


phillmv:
I double checked the versions and I think this is correct. Thanks @greysteil!

PR #384, opened 2019-03-28
Relax requirements for ActiveJob in CVE 2018-16476

mjankowski:
With the release of Rails 5.1.7 this advisory is coming up as a vulnerability, even though the fix is in place in 5.1.7.

This change relaxes the requirement to include versions greater than but outside of the 5.1.6.x series.


aselder:
👍 please

mjankowski:
Ah good catch, I'll revise.

mjankowski:
Updated to add 5.1.7 and leave the 5.1.6 line in place.

reedloden:
Makes sense. Thanks!

PR #390, opened 2019-06-02
Add CVE-2015-9284 for OmniAuth

GUI:
For a CSRF vulnerability in OmniAuth: https://github.com/omniauth/omniauth/pull/809

There's currently no patched version of OmniAuth to address this vulnerability, but there is an additional [omniauth-rails_csrf_protection gem](https://github.com/cookpad/omniauth-rails_csrf_protection) that patches the vulnerability if OmniAuth is being used with Rails. So I added a related link to this gem, but I'm open to suggestions for other ways to clarify this. See https://github.com/omniauth/omniauth/pull/809#issuecomment-497376674 and https://github.com/omniauth/omniauth/pull/809#issuecomment-497462504 for context on this separate gem and why it's only applicable to Rails usage.

### DIFF
--- a/gems/omniauth/CVE-2015-9284.yml
+++ b/gems/omniauth/CVE-2015-9284.yml
@@ -1,0 +1,19 @@
+---
+gem: omniauth
+cve: 2015-9284
+url: https://github.com/omniauth/omniauth/pull/809
+title: Cross-site request forgery (CSRF) vulnerability in OmniAuth's request phase
+date: 2015-05-25
+
+description: |
+  The request phase in OmniAuth is currently vulnerable to Cross-Site Request
+  Forgery, which allows an attacker to easily gain full access to a user's
+  account on a site that uses OmniAuth, when used in combination with another
+  CSRF vulnerability on the side of a connected OAuth provider.
+
+cvss_v2: 6.8
+cvss_v3: 8.8
+
+related:
+  url:
+    - https://github.com/cookpad/omniauth-rails_csrf_protection
--- a/gems/omniauth/CVE-2015-9284.yml
+++ b/gems/omniauth/CVE-2015-9284.yml
@@ -2,14 +2,20 @@
 gem: omniauth
 cve: 2015-9284
 url: https://github.com/omniauth/omniauth/pull/809
-title: Cross-site request forgery (CSRF) vulnerability in OmniAuth's request phase
+title: CSRF vulnerability in OmniAuth's request phase
 date: 2015-05-25
 
 description: |
-  The request phase in OmniAuth is currently vulnerable to Cross-Site Request
-  Forgery, which allows an attacker to easily gain full access to a user's
-  account on a site that uses OmniAuth, when used in combination with another
-  CSRF vulnerability on the side of a connected OAuth provider.
+  The request phase of the OmniAuth Ruby gem is vulnerable to Cross-Site
+  Request Forgery (CSRF) when used as part of the Ruby on Rails framework, allowing
+  accounts to be connected without user intent, user interaction, or feedback to
+  the user. This permits a secondary account to be able to sign into the web
+  application as the primary account.
+  
+  In order to mitigate this vulnerability, Rails users should consider using the
+  `omniauth-rails_csrf_protection` gem.
+  
+  More info is available here: https://github.com/omniauth/omniauth/pull/809#issuecomment-502079405
 
 cvss_v2: 6.8
 cvss_v3: 8.8


reedloden:
I was purposely waiting until upstream had a fix available... Would love to see the community's thoughts on continued waiting vs. publishing to the database without a fixed version.

BookOfGreg:
I think the CVE is actually misplaced. It's up to the implementors of the strategy to implement it correctly.
If this were merged into Rubysec, I'd simply have to disable it as no fix will be likely in Omniauth, but could be in Omniauth-rails or Devise and I'm not sure if this repo has the capability to say one gem fixes another. That's just my opinion and I could be wrong.

For me it's a Wait please.

GUI:
I guess from my perspective, if there's a known, public vulnerability or CVE issued, I'd like to know about it as soon as possible (even if there is no fix), so I can then decide what level of risk it poses to my own application and how I might want to address it.

While I do agree this specific case is a bit odd (since it's fuzzy where the responsibility for the fix lies), a default omniauth installation is theoretically vulnerable, so it seems like this could useful for users to know about. And while there's not a simple patched version of OmniAuth available, it seems like Rails users could benefit from knowing about omniauth-rails_csrf_protection (which addresses the underlying issue), and non-Rails users could also benefit from awareness, so they could manually address the issue if they deem it important enough (by setting `OmniAuth.config.allowed_request_methods = [:post]`, and then implementing a `OmniAuth.config.before_request_phase` that implements CSRF protection however their application implements it).

grepsedawk:
My 2 cents is that it's a misplaced vulnerability from mis-use of omniauth and the threat model is low.
Omniauth likely won't fix it because it's an implementation vuln...
This got kicked around for nearly 4 years before "becoming a CVE" and I'm still not convinced it's a vuln within the gem's responsibility.

Bundler audit (and other things using the ruby advisory db) is used more to say "hey there's an issue with this gem" while other tools, like brakeman, suggest that the way people implemented a feature could be vulnerable.

I could go either way here, but my vote would be that this doesn't go into the ruby-advisory-db without a fix unless this is specifically something that is a VERY serious issue. Like if a gem had a vuln and nobody *wants* to fix it, not "Nobody believes it's an issue with the gem"

respire:
I think this repo is "A database of vulnerable Ruby Gems".
since omniauth, a gem,  is vulnerable to CSRF attack, and it doesn't contain any warning or interface to prevent this attack, we should place this gem in the database.

however, this priority of this CVE is not that high IMO. It's okay to wait upstream for 1~2 months. If nothing got fixed after that, we can publish this CVE with a recommendation. for example, as GUI said, implement before_request_phase and change allowed_request_methods properly which is something similar to Workarounds in ruby / rails official security reports. 

Below: [CVE-2019-5418] Amendment: Possible Remote Code Execution Exploit in Action View
Workarounds

This vulnerability can be mitigated by specifying a format for file rendering,
like this:

```
class UserController < ApplicationController
  def index
    render file: "#{Rails.root}/some/file", formats: [:html]
  end
end
```

In summary, impacted calls to `render` look like this:

```
render file: "#{Rails.root}/some/file"
```

The vulnerability can be mitigated by changing to this:

```
render file: "#{Rails.root}/some/file", formats: [:html]
```

reedloden:
It's been over a month and still no fix upstream. Going to land this with a few modifications to the text.

Envek:
Okay, and how we should now get rid of that message?
```
$ bundle-audit check
…
Title: CSRF vulnerability in OmniAuth's request phase
Solution: remove or disable this gem until a patch is available!
```
Because there is no fix expected in omniauth gem itself as possible fixes are all implementation dependent (will be different for different frameworks), so in case of Rails we need to add separate `omniauth-rails_csrf_protection` gem (which depends on omniauth), in case of Sinatra we should do something other, in case of Hanami solution will be different too.

But omniauth neither will be removed after fix, nor updated.

I think that this advisory in its current form attracts attention to the problem (which is good), but isn't correctly bound to the omniauth gem itself.

Envek:
Currently we have to ignore this CVE after implementing fixes as @sikachu has proposed in https://github.com/omniauth/omniauth/pull/809#issuecomment-509098302 .
```sh
bundle audit check --ignore CVE-2015-9284
```

sikachu:
Hello, I'm the author of `omniauth-rails_csrf_protection` gem, and coming over after realized our CI is failing due to this PR (as `bundle audit` uses this database as a source). 😅

Sorry that I'm chiming in about this issue about 2 days too late. I totally forgot to check ruby-advisory-db if there's a CVE file pending to be added.

I agree with @Envek  — the fix for this issue will be different depending on framework that people are using, and it's more of application's configuration, so there's no "one size fit all" fix that can be included in future version of OmniAuth. That's why `omniauth-rails_csrf_protection` explicitly include "rails" in the gem name, as my fix only work if you're using Rails ...

Since this path pretty much affecting everyone who uses OmniAuth, and probably break their builds, would it be okay if I propose this solution?

1. Revert this PR temporary to make everyone's build temporary passing.
2. Add support for, say, `.bundler-audit.yml` in the project root so people can add CVE number that they have _mitigated_ or the one that they would like to _ignore_.
3. Add this CVE back, with a wording that "you can add CVE-2015-9284 to `.bundler-audit.yml` if you have mitigated the issue".

Please let me know what you think. I'll be happy to submit a patch to implement 2 and 3.

NobodysNightmare:
> Add this CVE back, with a wording that "you can add CVE-2015-9284 to .bundler-audit.yml if you have mitigated the issue".

I understand the idea here. But isn't this opening pandoras box? This could mean that at some point there is always some boilerplate "I took care of all the defaults" initialization to using `bundle-audit` (or any other tool depending on the advisory db). This could even include things like "if you don't go by the documentation, it will be insecure".

Don't get me wrong, I understand that proper security is oftentimes not as easy as just updating to a newer version. On the other hand, I usually do not get a warning for cases of insecurely using a secure tool.

Example: Recently Bootstrap fixed a XSS vulnerability by introducing a `sanitize` flag, which would default to `true`. All versions including this flag are now reported as not having a known vulernability anymore, regardless of users caring to simply pass `sanitize: false` and effectively disabling the protection again (maybe not all of them checking whether they pass user input).

Don't we have a similar case for Omniauth here? There is now a documented way to use the gem right, but the user is still free to use it wrong?

My main fear is that required "consenting" ("I read and mitigated all the well known issues") while initially setting up tools that depend on the advisory db, will decrease its adoption, which could be a bad thing as well. Imagine GitHub not effectively being able to scan public repositories anymore, because "normally it just fails because you forgot to add this YAML file".

mmrwoods:
> 2\. Add support for, say, `.bundler-audit.yml` in the project root so people can add CVE number that they have _mitigated_ or the one that they would like to _ignore_.

I'd also like to see support for some kind of ignore configuration added to bundler audit. 

For now, I've worked around the lack of support for this on one of our projects by adding advisories to a `config/bundler-audit.ignore` file, one advisory per line, commented lines starting with `#`, blank lines ignored.

We then use grep/cut/xargs to send the list of advisories as arguments when running `bundle audit` on CI, e.g. `egrep "^[^#]" config/bundle-audit.ignore | cut -f1 | xargs bundle audit check --update --ignore`

FWIW, here is an example file...

```
# Custom ignore file for bundler audit
#
# We sometimes need to ignore vulnerabilities in dependencies that we cannot
# easily upgrade, and are not exploitable or have been manually worked around.
#
# Bundler audit supports ignoring certain advisories using the --ignore option
# when checking, but, unlike Brakeman, it does not support the use of an ignore
# file. However, we really, really want to keep track of which advisories are
# being ignored and why, and we want that tracked in the code, and in git, so
# I'm using this custom file to store a list of advisories to be ignored. The
# file will be used to generate the input to the --ignore option when running
# bundler audit on jenkins, using the following very crude command:
#
# egrep "^[^#]" config/bundle-audit.ignore | cut -f1 | xargs bundle audit check --update --ignore
#
# Add one advisory per line, always add a comment explaining why an advisory is
# being ignored, and summarise the same in the commit that includes the change.

# Ignore CSRF vulnerability is omniauth
# Not exploitable because we've added omniauth-rails_csrf_protection to the gem
# bundle and updated admin authentication to post a form with authenticity_token
CVE-2015-9284
```

Envek:
Moreover, this vulnerability is application-specific. Only applications that implement authentication via multiple identity providers to the single application's account are vulnerable.

I believe that our application isn't vulnerable, because we don't have any “our own” accounts, we don't have logins and passwords, and we implement only one identity provider. So we have only “Sign in via eBay” button and nothing else as we're just a convenient tool for eBay users.

> In this scenario, the GitLab account would be linked to the attacker's Google account

In our case it is just impossible because if someone sign in to our app with another eBay account, we will authenticate them to another internal account in our app (or will create brand new one). Because we have strict 1-1 link between eBay accounts and our records in database.

olliebennett:
@mmrwoods thanks for your 'ignore file' implementation.

An alternative for ignoring a specific CVE is to pass the following flag;

```
bundle exec bundle-audit --ignore CVE-2015-9284
```

Issue #416, opened 2019-09-27
Include warnings for EOL ruby and gems 


cyc115:
End of life Ruby and Gems could be something `ruby-advisory-db` tracks. Tools like `bundler-audit` could then use this information to alert users fail builds. Any thoughts?

reedloden:
I like the idea. Thoughts on how we might show this to folks? Or the schema format?

BookOfGreg:
I like the idea, though can I ask it be the default but disable-able behind a flag. Since this forms part of CI for a lot of places this could end up blocking a lot of builds. Thoughts go out to a team I know contractually obligated to stay on ruby 1.8.x.

cyc115:
> I like the idea, though can I ask it be the default but disable-able behind a flag. Since this forms part of CI for a lot of places this could end up blocking a lot of builds. Thoughts go out to a team I know contractually obligated to stay on ruby 1.8.x.

Good idea, do you guys prefer to have it switch on with a flag (vs. on by default and switch off with flag)? Like you said, having it on by default could end up blocking a lot of builds.

postmodern:
End-of-Lifed rubies could be stored in [ruby-versions](https://github.com/postmodern/ruby-versions). End-of-Lifing isn't really a Security Advisory, so I don't think it really fits here. Although, Advisories for vulnerabilities in EOLed Rubies definitely can be added to ruby-advisory-db.

postmodern:
Closing this as the scope of [ruby-advisory-db](https://github.com/rubysec/ruby-advisory-db) is security advisories for vulnerabilities.

jasnow:
Might be of interest: https://github.com/marketplace/actions/xeol-end-of-life-eol-scan

PR #419, opened 2019-10-22
[RFC]: Support EOL

fatkodima:
#416 
I have included only 2 examples of how it can be implemented. 
`gems/eol.yml` file format:
```yaml
<gem name>:
  url:
  date:
  description:
```
I see `title` attribute here as always useless, because it will every time be the same, something like `"<gem name>  is deprecated"`.

`rubies/eol.yml` file format:
```yaml
<engine>:
  url:
  date:
  version:
```
Here I see `description` field as useless, as it will always be the same, like `"<engine> is deprecated, you should upgrade"`.

But for consistency, we can provide `title` and `description` in both variants.

I can think of an alternative implementation:
```
ruby-advisory-db/:
  eols/:
    gems/:
      ruby-sass.yml
      foo-bar.yml
      ...
   rubies/:
     ruby.yml
     jruby.yml
     ...
```

Where `ruby-sass.yml` has format:
```yaml
url:
date:
description:
```

And `ruby.yml` has format:
```yaml
url:
date:
version:
```

Will add missing tests after agreeing on implementation.

### DIFF
--- a/gems/eol.yml
+++ b/gems/eol.yml
@@ -1,0 +1,11 @@
+---
+ruby-sass:
+  url: https://github.com/sass/ruby-sass/issues/112
+  date: 2019-03-26
+
+  description: With the release of Dart Sass 1.0.0 stable, Ruby Sass was
+    officially deprecated. Applications should move to Dart Sass or LibSass.
+
+  related:
+    url:
+      - http://sass.logdown.com/posts/7081811
--- a/rubies/eol.yml
+++ b/rubies/eol.yml
@@ -1,0 +1,5 @@
+---
+ruby:
+  url: https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/
+  date: 2019-03-31
+  version: "< 2.4"


reedloden:
Have you considered `ruby-advisory-db/gems/ruby-saas/eol.yml` as a possible file location? That way, we keep all the old advisories in the same directory, but we also have a special `eol.yml` file in a gem directory that can describe that the gem is EOL.

fatkodima:
Sure, I have considered that option in the first place, but I'm a little worried about special case handling which will be introduced. At least in tests https://github.com/rubysec/ruby-advisory-db/blob/75e23aea6e5adf46d91fdb230810b0fcc2d47cf7/spec/advisories_spec.rb#L6-L21 and in `bundler-audit` in `Database` class in several places, like here https://github.com/rubysec/bundler-audit/blob/ad5afdadf6b792eca6e4ae872e1bb858fa9cdc2c/lib/bundler/audit/database.rb#L226-L228

And what do you think about .yml file structure?

postmodern:
If a gem/ruby is EOLed, but it does not have any security issues, why should you stop using it? I don't see why ruby-advisory-db should track EOLs. Seems a little out of scope.

postmodern:
Actually, would there be interest in creating a separate repo just for EOLs?

reedloden:
Personally, I consider the use of EOL software to be a security issue,
whether it has an actual security vulnerability or not. It means nobody is
actively maintaining the code, so if an issue is found, it may not ever be
fixed. Much easier to plan a migration away while there isn’t currently a
vulnerability than trying to rush something in the heat of the moment. In
fact, I’ve seen such a requirement to not use EOL software come up in
corporate legal agreements more than a few times.

postmodern:
@reedloden key word there is _when_ a security advisory is published after the EOL. The time between EOL and first advisory is a grey area.

I am still in favor of creating another repo database dedicated to EOLs or orphaned gems.

eoinkelly:
Gems sometimes go out of and then back into maintenance.  I think having one "EOL date" might not capture that easily. Would adding something like `maintained: true|false` to the schema be useful here?

Keithbreadley (approved):

postmodern:
I noticed this PR was approved, but still think it's out of scope of tracking security advisories. The amount of data to track whether a gem is currently maintained is much less than what is necessary to represent an advisory. Additionally, maybe this could be an attribute which could be added to [rubygems.org](https://github.com/rubygems/rubygems.org)'s database scheme?

simi:
I do suggest to close this as out of scope. Indeed there are better places to check for project status.

jasnow:
> I do suggest to close this as out of scope. Indeed there are better places to check for project status.

I am okay closing this PR - @fatkodima can create an update PR if they wish.

PR #421, opened 2019-10-22
Validate advisories through schemas

### LINKED ISSUE #370: No failed tests on unknown yaml structure
We have some great validations for well-formed .yml files.

One way to take them a step further is to validate each key in the .yml file is known/expected and fail otherwise.

With such tests, we'd have caught 25eb46646c70315ee35d580d8be0af2e8db283a6 before it landed.

An oldie, but a goodie, this project seems to allow you to easily define YML schema and have tests for it's well-formed nature: http://www.kuwata-lab.com/kwalify/

fatkodima:
Closes #370

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 gem 'faraday'
 gem 'rake'
+gem 'kwalify'
 gem 'rspec'
 
 group :development do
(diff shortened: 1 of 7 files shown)


reedloden:
Awesome! I see a few more changes, especially with regards to `ghsa`, but I can add those separately.

PR #430, opened 2020-02-05
Update CVE-2018-16396 and CVE-2018-16395 for Ruby 2.7.0

### LINKED ISSUE #429: Missing patched version ruby 2.7.0 for CVE-2018-16396 and CVE-2018-16395?
Hi,

when I use `ruby-audit` after upgrading ruby version to 2.7.0 I get two found vulnerabilities:
```
Name: ruby
Version: 2.7.0.0
Advisory: CVE-2018-16396
Criticality: Unknown
URL: https://www.ruby-lang.org/en/news/2018/10/17/not-propagated-taint-flag-in-some-formats-of-pack-cve-2018-16396/
Title: Tainted flags not always propogated in Array#pack and String#unpack
Solution: upgrade to ~> 2.3.8, ~> 2.4.5, ~> 2.5.2, ~> 2.6.0.pre.preview3

Name: ruby
Version: 2.7.0.0
Advisory: CVE-2018-16395
Criticality: Unknown
URL: https://www.ruby-lang.org/en/news/2018/10/17/openssl-x509-name-equality-check-does-not-work-correctly-cve-2018-16395/
Title: Incorrect equality check in OpenSSL::X509::Name
Solution: upgrade to ~> 2.3.8, ~> 2.4.5, ~> 2.5.2, ~> 2.6.0.pre.preview3
```
I don't have the expertise to decide whether the vulnerabilites are really still resolved in 2.7.0 but I would think so.

Is there any reason that the last line of the files
* https://github.com/rubysec/ruby-advisory-db/blob/master/rubies/ruby/CVE-2018-16396.yml
* https://github.com/rubysec/ruby-advisory-db/blob/master/rubies/ruby/CVE-2018-16395.yml

is 
```
~> 2.6.0.pre.preview3
```
 and not
```
>= 2.6.0.pre.preview3
```
(like in almost every other file)?

Jay-Schneider:
As discussed in #429 it seems to be a mistake that ruby versions `>= 2.7.0` are not listed as patched / unaffected version for those two CVEs:
1. [CVE-2018-16395](https://www.ruby-lang.org/en/news/2018/10/17/openssl-x509-name-equality-check-does-not-work-correctly-cve-2018-16395/)
which affects the openssl library which should not suddenly have become vulnerable again
1. [CVE-2018-16396](https://www.ruby-lang.org/en/news/2018/10/17/not-propagated-taint-flag-in-some-formats-of-pack-cve-2018-16396/)
which is specced to be fixed here: https://github.com/ruby/ruby/blob/master/spec/ruby/security/cve_2018_16396_spec.rb
 
resolves #429


phillmv:
Beg pardon, strictly speaking wouldn't it be `>= 2.6.0-preview3"`? Assuming there are no problems with that version requirement string.

Jay-Schneider:
No I don't think there is an issue with that suggestion. 

In my understanding it is just the same, isn't it? Or is there some catch about string additions to patch versions?

I will change the notation if you would prefer that for the sake of shortness.

phillmv:
>In my understanding it is just the same, isn't it? Or is there some catch about string additions to patch versions?

I forget the exact semantics of the `~>` operator.

If `~> 2.6.0-preview3` does that mean `2.6.1` is considered patched? In a strict since, it's versions after 2.6.0-preview3, and so a `>=` there should avoid these questions 🙂.

Jay-Schneider:
Hey @phillmv I get you, there is no reason to split the definition if it just requires more thinking to figure out what it means :)

Imo both notions should be equivalent but it does not really matter to me. I believe I just thought one line per minor version looks good but actually it is just unnecessary.

I updated this branch. You may re-review 🙂

reedloden:
Sorry for the massive delay with this.

reedloden:
Thank you so much, @Jay-Schneider

Issue #435, opened 2020-03-01
Add puma CVE-2020-5247


transoceanic2000:
HTTP Response splitting CVE in Puma - which is widely used (e.g. people running Rails on Heroku).

PR coming up.

reedloden:
Fixed by #436

PR #436, opened 2020-03-01
Add CVE-2020-5247 against puma, fixing #435

transoceanic2000:
NB: Patched versions are an extra point release higher than the CVE is declared against, as a bug was fixed against the immediate release and a further vector (early hints) was closed.


transoceanic2000:
@phillmv @reedloden Could one of you review this, and if appropriate merge it for me please.

reedloden:
Thank you for helping keep ruby users safe!

PR #443, opened 2020-04-15
Update vulns with patched versions

lloeki:
Some vulnerabilities have been patched in new releases. This avoids false positives.

Regarding `mapbox-rails` the versions are from the upstream JS library, but the gem versions match the upstream JS ones.

### DIFF
--- a/gems/fileutils/CVE-2013-2516.yml
+++ b/gems/fileutils/CVE-2013-2516.yml
@@ -1,4 +1,4 @@
---- 
+---
 gem: fileutils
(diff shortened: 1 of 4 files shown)


reedloden:
@lloeki Thanks for sending this in! I provided a few comments to address. Once corrected, would be happy to merge this.

lloeki:
@reedloden Thanks, I addressed the comments!

reedloden:
Awesome, thanks!

Issue #251, opened 2016-04-07
Migrate off of OSVDB


postmodern:
[OSVDB is shutting down](https://blog.osvdb.org/2016/04/05/osvdb-fin/). It's not clear what will become of OSVDB, whether it might be pivoted, change ownership or if we'll get a database dump. Currently, ruby-advisory-db contains OSVDB IDs and links to osvdb.org. The osvdb.org links should be replaced with other links, since they're all [404](http://www.osvdb.org/show/osvdb/98629)ed. We should probably also phase out the `OSVDB-...` file naming convention and remove osvdb references from the docs/examples to avoid confusing new contributors. The `osvdb:` fields should probably be removed at a later date, just in case OSVDB comes back to life.

postmodern:
IANACL (I am not a copyright lawyer), but can or should we remove the OSVDB legal notice from the README?

> This database also includes data from the Open Source Vulnerability Database
developed by the Open Security Foundation (OSF) and its contributors.

postmodern:
I propose we rename all `OSVDB-` files to `CVE-`, IFF they have a CVE ID. Advisory URLs should point back to the authoritative advisory from the project maintainers. This would be easy if we had a dump of OSVDB's database. We could point all advisory URLs to NVDB/MITRE in the interim?

connorshea:
Any updates on how the OSVDB is handling this shut down?

david-a-wheeler:
OSVDB is now gone; the front page should stop implying that it exists.  Where possible, a CVE identifier should be used in addition to the OSVDB identifier.  I don't see a problem with using both, to prevent future problems, but identifiers to defunct databases are less useful.

mveytsman:
I'm in the process of working on this.
- [ ] Find out if there are CVEs for anything we don't have them for
- [ ] Get CVEs issued for everything that only has OSVDB
- [ ] Replace all the URLs with non-osvdb links
- [ ] delete all OSVDB references in the yaml
- [ ] Update documentation & tests

FionaDL:
Hello @postmodern, @mveytsman I am interested in working on this issue.
I was thinking of replacing osvdb.org urls with nvd.nist.goc urls for all the current advisories, an example being: https://nvd.nist.gov/vuln/detail/CVE-2014-0130 .
Also I see there was tooling around osvdb.org in https://github.com/rubysec/ruby-advisory-db/blob/master/lib/osvdb_scrape.rb
Is this something that is worth to update?
[...]

Issue #441, opened 2020-04-02
Update titles/descriptions for NVD advisories.


FionaDL:
As suggested by @postmodern in comment https://github.com/rubysec/ruby-advisory-db/issues/251#issuecomment-606172546
it would be nice to update the titles/descriptions for each advisory on using NVD as a url. However, one thing I noticed is that there are no titles on NVD site. The descriptions could be updated. What do you think is the best approach @postmodern ?

postmodern:
Hmm let's wait on `titles` since that will require manually summarizing the description. Descriptions can be scraped from NVD (`//p[@data-testid="vuln-description"]`).

FionaDL:
Hi @simar7 !  Yes, I think the idea was that since the titles are leftover from the old links to the OSVDB website that we should update the titles and descriptions to match the links that they are now pointing too. Unfortunately the https://nvd.nist.gov site doesn't seem to use titles. They do have updated descriptions though.

simar7:
> Hi @simar7 ! Yes, I think the idea was that since the titles are leftover from the old links to the OSVDB website that we should update the titles and descriptions to match the links that they are now pointing at. Unfortunately the https://nvd.nist.gov site doesn't seem to use titles. They do have updated descriptions though.

Thanks. I've created a PR here to address it https://github.com/rubysec/ruby-advisory-db/pull/456

postmodern:
Closing since #456 was merged.

PR #449, opened 2020-06-12
chore: remove osvdb reference from LICENSE.txt

kgeorgiou:
Removes the osvdb reference in the License as a follow-up to https://github.com/rubysec/ruby-advisory-db/issues/251

Let me know if that makes sense - or not.

On a sidenote, might want to consider one of the common [OSS licenses](https://opensource.org/licenses) that GitHub formats nicely ([Example 1](https://github.com/FriendsOfPHP/security-advisories/blob/master/LICENSE), [Example 2](https://github.com/nodejs/security-wg/blob/master/LICENSE.md))


reedloden:
Unfortunately, until we remove all remnants of OSVDB data from the repository, this needs to stay. Once https://github.com/rubysec/ruby-advisory-db/issues/441 is done, we can do this.

Closing for now, but happy to reopen and merge once all uses of OSVDB data is removed.

PR #451, opened 2020-08-04
Added CVE-2020-16253 for pghero

ankane:
For https://github.com/ankane/pghero/issues/330


reedloden:
Thank you so much, @ankane!

PR #455, opened 2020-08-22
Link to dedicated omniauth page for CVE-2015-9284

ghiculescu:
https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284 is a good summary of the issue.


reedloden:
Thank you, @ghiculescu!

PR #456, opened 2020-08-25
Replace titles and descriptions with vuln-list advisories

simar7:
Since OSVDB is no longer maintained we would like to use
advisory information from an alternative source. RedHat provides
titles and descriptions for NVD CVEs as part of their feed.

vuln-list maintains the data from RedHat and other sources.
https://github.com/aquasecurity/vuln-list it is used by Trivy.

This commit replaces the titles and descriptions from the ones
that are maintained by vuln-list.

Signed-off-by: Simarpreet Singh <simar@linux.com>


reedloden:
Seems like your YML parser has broken a lot of the things (note the failing tests). Can you please modify your script to maintain the prevailing style? Alternatively, if you share your script, I am happy to take a look.

simar7:
> Seems like your YML parser has broken a lot of the things (note the failing tests). Can you please modify your script to maintain the prevailing style? Alternatively, if you share your script, I am happy to take a look.

Thanks for the feedback. Sorry for the late response, I was on vacation. I've addressed the failures.

knqyf263:
Hi @reedloden, which library do you use to generate these YAML advisories? I found the `yaml` library in GHSA sync, but I didn't find the library for other advisories.
https://github.com/rubysec/ruby-advisory-db/blob/1a7e4173860a83a432a4dc70676d3bce78e53a68/lib/github_advisory_sync.rb#L3

I think @simar7 used the official yaml module, but it looks like there are some differences such as patched_versions and affected_versions.
https://ruby-doc.org/stdlib-2.5.1/libdoc/yaml/rdoc/YAML.html

It is fine if you don't care about it since they are just ordering, an empty line, and etc. It doesn't change the data itself, but format.

reedloden:
Sorry for the long delay, @simar7.

The major issue is that I don't agree with the YAML changes in the current version, so that's why this hasn't been merged. We have a consistent format across all our files, and I don't see any reason why it should be made more complex. The YAML parser being used writes out YAML that is far more machine-parsable than human-readable. The latter is quite important for us, so I'd like to keep it.

Can you post the script you use? I'll see if I can just fix it to help move this along.

simar7:
> Sorry for the long delay, @simar7.
> 
> The major issue is that I don't agree with the YAML changes in the current version, so that's why this hasn't been merged. We have a consistent format across all our files, and I don't see any reason why it should be made more complex. The YAML parser being used writes out YAML that is far more machine-parsable than human-readable. The latter is quite important for us, so I'd like to keep it.
> 
> Can you post the script you use? I'll see if I can just fix it to help move this along.

Here's the script https://gist.github.com/simar7/99618d93fdc1fe0361a40392bceebfb7 - it fetches the info from vuln-list project which you'll need to clone locally and point the `rhRoot` variable to before running it. You can place this script within this repository to run it. You can get vuln-list project from here https://github.com/aquasecurity/vuln-list.

IMO after the initial review I fixed the YAML output to be as close as it is to the original. In fact I don't see any changes except ordering of a few items and some empty lines. But maybe I have missed something obvious.

reedloden:
> IMO after the initial review I fixed the YAML output to be as close as it is to the original. In fact I don't see any changes except ordering of a few items and some empty lines. But maybe I have missed something obvious.

Is it possible you didn't push that new update up to your branch? The version I currently see in this PR removes the `---` at the top of all the files, as well as adds unnecessary quotes and collapses multi-line things into a single-line. Makes it pretty unreadable currently.

simar7:
> > IMO after the initial review I fixed the YAML output to be as close as it is to the original. In fact I don't see any changes except ordering of a few items and some empty lines. But maybe I have missed something obvious.
> 
> Is it possible you didn't push that new update up to your branch? The version I currently see in this PR removes the `---` at the top of all the files, as well as adds unnecessary quotes and collapses multi-line things into a single-line. Makes it pretty unreadable currently.

I have pushed up all the changes I had. Unfortunately I don't have cycles at the moment to work on this but happy to let someone else to continue from what I have here.

reedloden (approved):
Thanks for all the work on this. I finally had time to use your script and clean things up. I rebased and pushed to your original branch.

PR #458, opened 2020-12-22
Remove misleading fileutils OSVDB reports

mensfeld:
As a followup to my discussion with Hiroshi Shibata, maintainer of the https://rubygems.org/gems/fileutils gem:

Zrzut ekranu z 2020-12-22 14-31-59

The following reports are no longer valid and completely misleading, causing the users of those to get false alarms in case fileutils ends up in the gemfile.


postmodern:
Judging by the dates and descriptions of the OSVDB advisories, they all correlate to CVE-2013-2516, which we already have imported.

PR #468, opened 2021-04-30
Add Bundler CVE-2020-36327

zofrex:
PR to add the CVE for implicit dependency confusion.

I have two questions about this:

1) Is it correct to put 2.2.10 in the `unaffected_versions` section? Technically that version was patched (and the patch was rolled back in 2.2.11), so `patched_versions` seems more appropriate. However, if I put it there then a test fails:

```
Failures:

  1) gems /Users/zofrex/Code/ruby/ruby-advisory-db/spec/../gems/bundler/CVE-2020-36327.yml versions assumes that future versions will be patched
     Failure/Error: expect(versions.last.match(/^>=|^>/)).to be_truthy
     
       expected: truthy value
            got: nil
     Shared Example Group: "Gem Advisory" called from ./spec/advisories_spec.rb:8
     # ./spec/gem_example.rb:32:in `block (4 levels) in <top (required)>'
```

2) I wrote a blog post about this vulnerability, explaining it in more detail and outlining some possible mitigations. Would it be OK to add the URL to the related section? I've left it out to start with in case self-promotion isn't ok.

### DIFF
--- a/gems/bundler/CVE-2020-36327.yml
+++ b/gems/bundler/CVE-2020-36327.yml
@@ -1,0 +1,33 @@
+---
+gem: bundler
+cve: 2020-36327
+ghsa: fp4w-jxhp-m23p
+date: 2020-09-30
+url: https://github.com/rubygems/rubygems/issues/3982
+title: Dependency Confusion in Bundler with Implicit Private Dependencies
+
+description: |
+  Bundler 1.16.0 through 2.2.9 and 2.2.11 through 2.2.17 sometimes chooses a
+  dependency source based on the highest gem version number, which means that a
+  rogue gem found at a public source may be chosen, even if the intended choice
+  was a private gem that is a dependency of another private gem that is
+  explicitly depended on by the application.
+
+cvss_v3: 8.8
+
+patched_versions:
+  - ">= 2.2.18"
+  - "2.2.10"
+
+unaffected_versions:
+  - "< 1.16.0"
+
+related:
+  cve:
+    - 2021-24105
+  url:
+    - https://bundler.io/blog/2021/02/15/a-more-secure-bundler-we-fixed-our-source-priorities.html
+    - https://mensfeld.pl/2021/02/rubygems-dependency-confusion-attack-side-of-things/
+    - https://www.zofrex.com/blog/2021/04/29/bundler-still-vulnerable-dependency-confusion-cve-2020-36327/
+    - https://msrc.microsoft.com/update-guide/en-US/vulnerability/CVE-2021-24105
+    - https://github.com/rubygems/rubygems/pull/4609


gems/bundler/CVE-2020-36327.yml
```diff
@@ -0,0 +1,27 @@
...
+  Spoofing
+
+description: |
+  Bundler 1.16.0 through 2.2.9 and 2.2.11 through 2.2.16 sometimes chooses a
+  dependency source based on the highest gem version number, which means that a
+  rogue gem found at a public source may be chosen, even if the intended choice
+  was a private gem that is a dependency of another private gem that is
+  explicitly depended on by the application.
+
+patched_versions:
```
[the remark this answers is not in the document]
zofrex: It was fixed in 2.2.10 but the fix was rolled back in 2.2.11. It remains unfixed in 2.2.17.

gems/bundler/CVE-2020-36327.yml
```diff
@@ -0,0 +1,27 @@
...
+unaffected_versions:
```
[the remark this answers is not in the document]
zofrex: The bug was introduced in 1.16.0 so I think that part is correct at least.

gems/bundler/CVE-2020-36327.yml
```diff
@@ -0,0 +1,20 @@
+---
+gem: bundler
+cve: 2020-36327
+date: 2020-09-30
+url: https://nvd.nist.gov/vuln/detail/CVE-2020-36327
```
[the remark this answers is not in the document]
zofrex: I agree, that's the best option here. Good call!

zofrex:
> So, 2.2.10 and 2.2.17 are both fixed? I'm just a bit confused about what versions have the fix in them.

Just 2.2.10 is fixed.

reedloden:
Are there plans for Bundler to fix this? I'm a bit hesitant to just add an entry that basically tells everybody to just install only Bundler v2.2.10, when the Bundler team reverted https://github.com/rubygems/rubygems/pull/4385 and still hasn't fixed it yet. Seems like https://github.com/rubygems/rubygems/pull/4381 may be a partial fix for this that was included in v2.2.14?

@deivid-rodriguez can you help here? Is this CVE fully addressed in some prior Bundler release after v2.2.10?

deivid-rodriguez:
We will release a fix for this imminently as bundler 2.2.18.

gems/bundler/CVE-2020-36327.yml
```diff
@@ -0,0 +1,28 @@
...
+  Bundler 1.16.0 through 2.2.9 and 2.2.11 through 2.2.16 sometimes chooses a
+  dependency source based on the highest gem version number, which means that a
+  rogue gem found at a public source may be chosen, even if the intended choice
+  was a private gem that is a dependency of another private gem that is
+  explicitly depended on by the application.
+
+patched_versions:
+unaffected_versions:
+  - "< 1.16.0"
+  - "2.2.10"
```
reedloden, line 20: Move to `patched_versions`

gems/bundler/CVE-2020-36327.yml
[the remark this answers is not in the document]
reedloden: Let's add `>= 2.2.18` in advance of the fixed bundler version.

gems/bundler/CVE-2020-36327.yml
```diff
@@ -0,0 +1,28 @@
...
+gem: bundler
+cve: 2020-36327
+date: 2020-09-30
+url: https://github.com/rubygems/rubygems/issues/3982
+title: |
+  Bundler Gem for Ruby Implicit Dependencies Gemfile Handling Gem Installation
+  Spoofing
+
+description: |
+  Bundler 1.16.0 through 2.2.9 and 2.2.11 through 2.2.16 sometimes chooses a
```
reedloden, line 11: Change 2.2.16 to 2.2.17. Looks like 2.2.18 will be the fixed version.

reedloden (changes requested):
I don't want to merge this until the fixed version of bundler is released, but let's just make sure it's ready to go before then.

reedloden:
Looks like GitHub released a security advisory for this as well.

So, let's add `ghsa: fp4w-jxhp-m23p` and `cvss_v3: 8.8`.

zofrex:
Happy to make the requested changes as and when the time is right, but I must disagree with waiting to merge this until a fix is out.

> I'm a bit hesitant to just add an entry that basically tells everybody to just install only Bundler v2.2.10

Really what this entry does is tell people "you might be vulnerable to this bug, and if you are, here are your options". I am definitely hesitant to push people towards 2.2.10 because that did have a lot of issues, but there are also several mitigations outlined in the links in this entry.

In my opinion an advisory without a patch is still hugely valuable to users because it lets us know there might be an issue, we can then look into whether or not it affects us, and if it does, make a decision about whether our personal risk exposure is high enough to apply one of the mitigations while we wait for the patch.

This isn't a hugely unusual scenario, advisories coming out with no patch happens from time to time, and I'd rather know about it than not. In extreme cases I have taken systems offline or avoided using them, in others applied a mitigation, and in others, crossed my fingers, but that's a decision that users can only make for themselves, and a decision that is best guided by having all the available information in front of them.

For that reason I strongly urge you to merge this as soon as possible. We can update it as the situation changes.

gems/bundler/CVE-2020-36327.yml
zofrex: I don't think < 1.16.0 is "patched" because that predates the vulnerability being introduced? 2.2.10 is but, as discussed above, that breaks the tests.

gems/bundler/CVE-2020-36327.yml
zofrex: Let's add ">= 2.2.18" when that's confirmed.

gems/bundler/CVE-2020-36327.yml
reedloden: I made a change, so I think the tests will work now.

reedloden (approved):

deivid-rodriguez:
Bundler 2.2.18 has been released.

zofrex:
Rebased and squashed, as I think this is ready to go now.

phillmv:
Merged! Thank you all for your work. If need be, fwiw, we can also find a way to have better sync/coordination with GitHub advisories.

PR #469, opened 2021-05-05
Add CVE-2021-28966 for tmpdir

ooooooo-q:
Here is the modified commit on the repository side of gem. https://github.com/ruby/tmpdir/commit/adf294bc2d10cd223aa3ca488079ec313032c07b


reedloden:
Thank you so much!

Issue #402, opened 2019-07-07
[GHSA sync] add extra references


reedloden:
Aside from the primary reference URL, should add any additional references under the `related` section.

reedloden:
@rschultheis ^^

jasnow:
This might be done.

postmodern:
This can be closed now. The GHSA sync script now populates `related: url: ...`.

jasnow:
Thanks

Issue #400, opened 2019-07-06
[GHSA sync] handled related IDs


reedloden:
Sometimes we have multiple IDs combined together into one advisory.

One example of this is `gems/bootstrap/CVE-2018-14040.yml`. The GHSA sync script doesn't understand that CVE-2018-14042 is part of this same advisory.

reedloden:
@rschultheis ^^

jasnow:
The idea behind this issue is for the GHSA sync script to combine multiple GHSA advisories into one (fewer) ruby-advisory-db advisories and use related:/cvs: field and related:/ghsa: field because they are the "same" advisory.

jasnow:
@rubysec/maintainers @reedloden @postmodern - I believe we have address this issue - Can we close this issue now?

Issue #399, opened 2019-07-06
[GHSA sync] Pull CVSS score from NVD


reedloden:
Can we pull the CVSS score(s) from NVD via https://github.com/olbat/nvdcve when writing advisories?

reedloden:
@rschultheis ^^

postmodern:
btw, I have plans to write an NVD library that we can use.

phillmv:
fun fact: we're actually in the process of backloading cvss and cwes into the gh data. maybe another month or so should be able to sync it off the github feed 🤔

jasnow:
Is this done?

postmodern:
Closing this as the `github_advisory_sync` script now populates CVSS scores.

PR #479, opened 2021-06-19
add missing cvss_v3 scores during github sync

SkipKayhil:
I came across some vulnerabilities that were missing cvss scores present on Github, so I wanted to update the github sync script to pull down those values automatically. I broke the change into three commits:
1. The first commit is a refactor of the github sync to query vulnerabilities instead of advisories. This was done because I noticed that the number of queries made to pull down all of the vulnerabilities seemed too high (43 when I first ran it), and it turned out that it was getting advisories for all ecosystems and not just ruby. After the refactor, it is able to pull all of the ruby vulnerabilities and advisories in 7 requests.
2. The second commit is the change that adds missing cvss scores if the score is present on Github.
3. The third is the result of running the script, adding 25 missing cvss scores

Let me know if you have any feedback or if there is anything I should do differently that could help this get merged!

### DIFF
--- a/lib/github_advisory_sync.rb
+++ b/lib/github_advisory_sync.rb
@@ -5,7 +5,6 @@ require "open-uri"
 
 module GitHub
   class GitHubAdvisorySync
-
     # Sync makes sure there are rubysec advisories for all GitHub advisories
     # It writes a set of yaml files, one for each GitHub Advisory that
     # is not already present in this repo
@@ -14,22 +13,12 @@ module GitHub
     # It is more important to sync the newer ones, so this allows the user to
     # control how old of CVEs the sync should pull over
     def self.sync(min_year: 2015)
-      gh_advisories = GraphQLAPIClient.new.retrieve_all_rubygem_publishable_advisories
+      gh_advisories = GraphQLAPIClient.new.all_rubygem_advisories
 
       # Filter out advisories with a CVE year that is before the min_year
-      gh_advisories.select! do |advisory|
-        if advisory.cve_id
-          _, cve_year = advisory.cve_id.match(/^CVE-(\d+)-\d+$/).to_a
-          cve_year.to_i >= min_year
-        else
-          true # all advisories without a CVE are included too
-        end
-      end
+      gh_advisories.select! { |v| v.cve_after_year?(min_year) }
 
-      files_written = []
-      gh_advisories.each do |advisory|
-        files_written += advisory.write_files
-      end
+      files_written = gh_advisories.filter_map(&:write_files).flatten.compact!
 
       puts "\nSync completed"
       if files_written.empty?
@@ -43,14 +32,15 @@ module GitHub
   end
 
   class GraphQLAPIClient
-    GITHUB_API_URL = "https://api.github.com/graphql"
+    GITHUB_API_URL = "https://api.github.com/graphql".freeze
 
     GitHubApiTokenMissingError = Class.new(StandardError)
 
     # return a lazy initialized connection to github api
     def github_api(adapter = :net_http)
-      @faraday_connection ||= begin
+      @github_api ||= begin
         puts "Initializing GitHub API connection to URL: #{GITHUB_API_URL}"
+
         Faraday.new do |conn_builder|
           conn_builder.adapter adapter
           conn_builder.headers = {
@@ -60,7 +50,6 @@ module GitHub
           }
         end
       end
-      @faraday_connection
     end
 
     # An error class which gets raised when a GraphQL request fails
@@ -70,54 +59,72 @@ module GitHub
     # error checking and how queries and requests are formed
     def github_graphql_query(graphql_query_name, graphql_variables = {})
       graphql_query_str = GraphQLQueries.const_get graphql_query_name
-      graphql_body = JSON.generate query: graphql_query_str,
-                                   variables: graphql_variables
+      graphql_body = JSON.generate(query: graphql_query_str, variables: graphql_variables)
+
       puts "Executing GraphQL request: #{graphql_query_name}. Request variables:\n#{graphql_variables.to_yaml}\n"
+
       faraday_response = github_api.post do |req|
         req.url GITHUB_API_URL
         req.body = graphql_body
       end
+
       puts "Got response code: #{faraday_response.status}"
+
       if faraday_response.status != 200
-        raise(GitHubGraphQLAPIError, "GitHub GraphQL request to #{faraday_response.env.url} failed: #{faraday_response.body}")
+        raise(
+          GitHubGraphQLAPIError,
+          "GitHub GraphQL request to #{faraday_response.env.url} failed: #{faraday_response.body}"
+        )
       end
+
       body_obj = JSON.parse faraday_response.body
+
       if body_obj["errors"]
         raise(GitHubGraphQLAPIError, body_obj["errors"].map { |e| e["message"] }.join(", "))
       end
+
       body_obj
     end
 
-    def retrieve_all_github_advisories(max_pages = 1000, page_size = 100)
-      all_advisories = []
-      variables = { "first" => page_size }
-      max_pages.times do |page_num|
-        puts "Getting page #{page_num + 1} of GitHub Advisories"
-        page = github_graphql_query(:GITHUB_ADVISORIES_WITH_RUBYGEM_VULNERABILITY, variables)
-        advisories_this_page = page["data"]["securityAdvisories"]["nodes"]
-        all_advisories += advisories_this_page
-        break unless page["data"]["securityAdvisories"]["pageInfo"]["hasNextPage"] == true
-        variables["after"] = page["data"]["securityAdvisories"]["pageInfo"]["endCursor"]
-      end
-      puts "Retrieved #{all_advisories.length} Advisories from GitHub API"
+    def all_rubygem_advisories
+      advisories = {}
 
-      all_advisories.map do |advisory_graphql_obj|
-        GitHubAdvisory.new github_advisory_graphql_object: advisory_graphql_obj
+      retrieve_all_rubygem_vulnerabilities.each do |vulnerability|
+        advisory = GitHubAdvisory.new(vulnerability["advisory"])
+
+        next if advisory.withdrawn?
+
+        advisories[advisory.primary_id] ||= advisory
+
+        advisories[advisory.primary_id].vulnerabilities << vulnerability.except("advisory")
       end
+
+      advisories.values
     end
 
-    def retrieve_all_rubygem_publishable_advisories
-      all_advisories = retrieve_all_github_advisories
-      # remove withdrawn advisories,
-      # and remove those where there are no vulnerabilities for ruby
-      all_advisories.reject { |advisory| advisory.withdrawn? }
-                    .select { |advisory| advisory.has_ruby_vulnerabilities? }
+    def retrieve_all_rubygem_vulnerabilities(max_pages = 1000, page_size = 100)
+      all_vulnerabilities = []
+      variables = { "first" => page_size }
+      max_pages.times do |page_num|
+        puts "Getting page #{page_num + 1} of GitHub Vulnerabilities"
+
+        page = github_graphql_query(:RUBYGEM_VULNERABILITIES_WITH_GITHUB_ADVISORIES, variables)
+        vulnerabilities_this_page = page["data"]["securityVulnerabilities"]["nodes"]
+        all_vulnerabilities += vulnerabilities_this_page
+
+        break unless page["data"]["securityVulnerabilities"]["pageInfo"]["hasNextPage"] == true
+
+        variables["after"] = page["data"]["securityVulnerabilities"]["pageInfo"]["endCursor"]
+      end
+      puts "Retrieved #{all_vulnerabilities.length} Vulnerabilities from GitHub API"
+
+      all_vulnerabilities
     end
 
     module GraphQLQueries
-      GITHUB_ADVISORIES_WITH_RUBYGEM_VULNERABILITY = <<-GRAPHQL.freeze
+      RUBYGEM_VULNERABILITIES_WITH_GITHUB_ADVISORIES = <<-GRAPHQL.freeze
         query($first: Int, $after: String) {
-          securityAdvisories(first: $first, after: $after) {
+          securityVulnerabilities(first: $first, after: $after, ecosystem:RUBYGEMS) {
             pageInfo {
               endCursor
               hasNextPage
@@ -125,33 +132,31 @@ module GitHub
               startCursor
             }
             nodes {
-              identifiers {
-                type
-                value
+              package {
+                name
+                ecosystem
               }
-              summary
-              description
-              severity
-              cvss {
-                score
-                vectorString
+              vulnerableVersionRange
+              firstPatchedVersion {
+                identifier
               }
-              references {
-                url
-              }
-              publishedAt
-              withdrawnAt
-              vulnerabilities(ecosystem:RUBYGEMS, first: 10) {
-                nodes {
-                  package {
-                    name
-                    ecosystem
-                  }
-                  vulnerableVersionRange
-                  firstPatchedVersion {
-                    identifier
-                  }
+              advisory {
+                identifiers {
+                  type
+                  value
                 }
+                summary
+                description
+                severity
+                cvss {
+                  score
+                  vectorString
+                }
+                references {
+                  url
+                }
+                publishedAt
+                withdrawnAt
               }
             }
           }
@@ -163,22 +168,26 @@ module GitHub
 
     def github_api_token
       unless ENV["GH_API_TOKEN"]
-        raise GitHubApiTokenMissingError, "Unable to make API requests.  Must define 'GH_API_TOKEN' environment variable."
+        raise(
+          GitHubApiTokenMissingError,
+          "Unable to make API requests.  Must define 'GH_API_TOKEN' environment variable."
+        )
       end
+
       ENV["GH_API_TOKEN"]
     end
   end
 
   class GitHubAdvisory
+    attr_reader :advisory, :vulnerabilities
 
-    attr_reader :github_advisory_graphql_object
-
-    def initialize(github_advisory_graphql_object:)
-      @github_advisory_graphql_object = github_advisory_graphql_object
+    def initialize(advisory)
+      @advisory = advisory
+      @vulnerabilities = []
     end
 
     def identifier_list
-      github_advisory_graphql_object["identifiers"]
+      advisory["identifiers"]
     end
 
     # extract the CVE identifier from the GitHub Advisory identifier list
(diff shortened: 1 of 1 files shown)


postmodern:
@reedloden or @rschultheis could you review the changes to the github sync script, there's a lot of changes.

lib/github_advisory_sync.rb
```diff
@@ -70,88 +59,102 @@ def github_api(adapter = :net_http)
...
+          "GitHub GraphQL request to #{faraday_response.env.url} failed: #{faraday_response.body}"
+        )
       end
+
       body_obj = JSON.parse faraday_response.body
-      if body_obj["errors"]
-        raise(GitHubGraphQLAPIError, body_obj["errors"].map { |e| e["message"] }.join(", "))
-      end
+
+      raise(GitHubGraphQLAPIError, body_obj["errors"].map { |e| e["message"] }.join(", ")) if body_obj["errors"]
```
phillmv, line 82: minor style issue:

for my money, better to have

```
if error
  raise
end
```

than `raise foo if error`, this raise line is almost trivial enough that `foo if errors` is OK but not quite trivial enough for my money ;)

lib/github_advisory_sync.rb
```diff
@@ -43,14 +32,15 @@ def self.sync(min_year: 2015)
...
   class GraphQLAPIClient
-    GITHUB_API_URL = "https://api.github.com/graphql"
+    GITHUB_API_URL = "https://api.github.com/graphql".freeze
 
     GitHubApiTokenMissingError = Class.new(StandardError)
 
     # return a lazy initialized connection to github api
     def github_api(adapter = :net_http)
-      @faraday_connection ||= begin
+      @github_api ||= begin
```
phillmv, line 41: :+1:

lib/github_advisory_sync.rb
```diff
@@ -70,88 +59,102 @@ def github_api(adapter = :net_http)
...
-      all_advisories = []
+    def all_rubygem_advisories
+      advisories = {}
+
+      retrieve_all_rubygem_vulnerabilities.each do |vulnerability|
+        advisory = GitHubAdvisory.new(vulnerability["advisory"])
+
+        next if advisory.withdrawn?
+
+        advisories[advisory.primary_id] = advisory unless advisories.key? advisory.primary_id
```
phillmv, line 95: i believe this will do just fine too 😄 

```suggestion
        advisories[advisory.primary_id] ||= advisory
```

lib/github_advisory_sync.rb
```diff
@@ -70,88 +59,102 @@ def github_api(adapter = :net_http)
...
-                    .select { |advisory| advisory.has_ruby_vulnerabilities? }
+      all_vulnerabilities
     end
 
     module GraphQLQueries
-      GITHUB_ADVISORIES_WITH_RUBYGEM_VULNERABILITY = <<-GRAPHQL.freeze
+      RUBYGEM_VULNERABILITIES_WITH_GITHUB_ADVISORIES = <<-GRAPHQL.freeze
         query($first: Int, $after: String) {
-          securityAdvisories(first: $first, after: $after) {
+          securityVulnerabilities(first: $first, after: $after, ecosystem:RUBYGEMS) {
```
phillmv, line 125: Very good! We'd recently moved ecosystem down to the `vulnerability` node, and I'm glad you noticed this here.

Seeing this tho, I wonder two things @SkipKayhil:

- life would be easier if offered an `ecosystem` filter on securityAdvisories, eh?
- would it be useful to provide an `updated_after` filter, so we don't have to page thru every single advisory? that way maybe we could just store a timestamp of the last time the importer was run.

lib/github_advisory_sync.rb
```diff
@@ -199,110 +206,122 @@ def ghsa_id
...
-      return "<FILL IN IF AVAILABLE>" if github_advisory_graphql_object["cvss"]["vectorString"].nil?
-      github_advisory_graphql_object["cvss"]["score"].to_f
+      return if advisory["cvss"]["vectorString"].nil?
+
+      advisory["cvss"]["score"].to_f
     end
 
     def external_reference
-      github_advisory_graphql_object["references"].find do |ref|
+      advisory["references"].each do |ref|
```
phillmv, line 233: this is fine but i would have left `find` in place.

phillmv: also of note is iirc we often include many different references now in the data, so looking for the first non-nvd url might not always be what we want 🤔

lib/github_advisory_sync.rb
```diff
@@ -14,22 +13,12 @@ class GitHubAdvisorySync
...
-          true # all advisories without a CVE are included too
-        end
-      end
+      gh_advisories.select! { |v| v.cve_after_year?(min_year) }
 
-      files_written = []
-      gh_advisories.each do |advisory|
-        files_written += advisory.write_files
-      end
+      files_written = gh_advisories.filter_map(&:write_file)
```
phillmv, line 21: i approve of this main loop simplification!

lib/github_advisory_sync.rb
```diff
@@ -163,22 +166,26 @@ module GraphQLQueries
...
   end
 
   class GitHubAdvisory
+    attr_reader :advisory, :vulnerabilities
 
-    attr_reader :github_advisory_graphql_object
-
-    def initialize(github_advisory_graphql_object:)
-      @github_advisory_graphql_object = github_advisory_graphql_object
+    def initialize(advisory)
```
phillmv, line 182: i also approve of this simplification!

lib/github_advisory_sync.rb
```diff
@@ -199,110 +206,122 @@ def ghsa_id
...
-        ref["url"]
+
+        return ref["url"]
       end
     end
 
-    def vulnerabilities
-      github_advisory_graphql_object["vulnerabilities"]["nodes"]
+    def package_name
+      vulnerabilities.first["package"]["name"]
```
phillmv, line 241: in practice, this is fine but of note is that it's technically possible for one advisory to span many different package names.

lib/github_advisory_sync.rb
```diff
@@ -199,110 +206,122 @@ def ghsa_id
...
-          #
-          # To help write those fields, we put all the github data below.
-          #
-          # The second block of yaml in a .yaml file is ignored (after the second "---" line)
-          # This effectively makes this data a large comment
-          # Still it should be removed before the data goes into rubysec
-          file.write "\n\n# GitHub advisory data below - **Remove this data before committing**\n"
-          file.write "# Use this data to write patched_versions (and potentially unaffected_versions) above\n"
-          file.write github_advisory_graphql_object.to_yaml
+      return if saved_data.key?("cvss_v3") || cvss.nil?
```
phillmv, line 251: if we're only checking to update cvss_v3, maybe we can rename the method accordingly, `update_file_cvss_v3` or something.

lib/github_advisory_sync.rb
```diff
@@ -199,110 +206,122 @@ def ghsa_id
...
+
+      cvss_added = false
+      File.open("#{rubysec_filename}.tmp", "w") do |f| 
+        IO.foreach(rubysec_filename) do |line|
+          if (line.include?('unaffected_versions:') || line.include?('patched_versions:')) && !cvss_added
```
phillmv, line 256: This gives me pause. Why not just add the cvss_v3 key to the saved_data, and re-emit the YAML wholesale? Is the diff too ugly otherwise?

phillmv (commented):
I like it. I read over the code, and it seems largely benign. I left a variety of comments, mostly minor style observations.

My recommendation is thus:

Can you split off the updates to the data into a 2nd, separate, PR so that we can look only at the changes to the `lib/github_advisory_sync`? Then once this is approved, we can just rubber-stamp the data update.

lib/github_advisory_sync.rb
SkipKayhil: Both of those sounds like great additions! I think adding ecosystem to `securityAdvisories` would be a huge win for sure, and probably remove a lot of the refactoring I did to invert the relationship

phillmv: We're probably going to touch the API this summer, so I'll add this to my list of tweaks. If you have any other suggestions, holler.

lib/github_advisory_sync.rb
SkipKayhil: I agree, this is pretty ugly. The issue with `YAML.load/dump` was that it wasn't dumping multi line strings for me, the new file would look like this:
```yaml
---
gem: actionpack
framework: rails
cve: 2015-7581
date: 2016-01-25
url: https://groups.google.com/forum/#!topic/rubyonrails-security/dthJ5wL69JE
title: Object leak vulnerability for wildcard controller routes in Action Pack
description: "There is an object leak vulnerability for wildcard controllers in Action
  Pack. \nThis vulnerability has been assigned the CVE identifier CVE-2015-7581. \n\nVersions
  Affected:  >= 4.0.0 and < 5.0.0.beta1 \nNot affected:       < 4.0.0, 5.0.0.beta1
  and newer \nFixed Versions:     4.2.5.1, 4.1.14.1 \n\nImpact \n------ \nUsers that
  have a route that contains the string \":controller\" are susceptible \nto objects
  being leaked globally which can lead to unbounded memory growth. \nTo identify if
  your application is vulnerable, look for routes that contain \n\":controller\".
  \n\nInternally, Action Pack keeps a map of \"url controller name\" to \"controller
  \nclass name\".  This map is cached globally, and is populated even if the \ncontroller
  class doesn't actually exist. \n\nAll users running an affected release should either
  upgrade or use one of the \nworkarounds immediately. \n\nReleases \n-------- \nThe
  FIXED releases are available at the normal locations. \n\nWorkarounds \n-----------
  \nThere are no feasible workarounds for this issue. \n\nPatches \n------- \nTo aid
  users who aren't able to upgrade immediately we have provided patches for the two
  supported release series.  They are in git-am format and consist of a single changeset.
  \n\n* 4-1-wildcard_route.patch - Patch for 4.1 series \n* 4-2-wildcard_route.patch
  - Patch for 4.2 series \n\nPlease note that only the 4.1.x and 4.2.x series are
  supported at present.  Users of earlier unsupported releases are advised to upgrade
  as soon as possible as we cannot guarantee the continued availability of security
  fixes for unsupported releases.\n"
unaffected_versions:
- "< 4.0.0"
- ">= 5.0.0.beta1"
patched_versions:
- "~> 4.2.5, >= 4.2.5.1"
- "~> 4.1.14, >= 4.1.14.1"
cvss_v3: 7.5
```

If I'm missing some parameter that would make it keep the "pretty" multi line strings then that would be way better than what I tried

lib/github_advisory_sync.rb
reedloden: Do we expect that to ever happen? If so, we should probably support that... See also https://github.com/rubysec/ruby-advisory-db/issues/400

lib/github_advisory_sync.rb
reedloden: Can we support updating other things? Would be nice to have an (optional) way to sync more than just CVSSv3, as if the advisories are updated, we might not pick up that at all.

lib/github_advisory_sync.rb
```diff
@@ -199,110 +208,120 @@ def ghsa_id
...
+        "url" => external_reference,
+        "date" => published_day,
+        "title" => advisory["summary"],
+        "description" => advisory["description"],
+        "cvss_v3" => cvss,
+        "patched_versions" => ["<FILL IN SEE BELOW>"],
+        "unaffected_versions" => ["<OPTIONAL: FILL IN SEE BELOW>"]
+      }
+      data["cve"] = cve_id[4..20] if cve_id
+      data["cvss_v3"] = "<FILL IN IF AVAILABLE>" if cvss.nil?
```
reedloden, line 287: Is this where we should at some future point use NVD's API to check if they have an available score? *cough* https://github.com/rubysec/ruby-advisory-db/issues/399

lib/github_advisory_sync.rb
```diff
@@ -199,110 +208,120 @@ def ghsa_id
...
+
+        # The data we just wrote is incomplete,
+        # and therefore should not be committed as is
+        # We can not directly translate from GitHub to rubysec advisory format
+        #
+        # The patched_versions field is not exactly available.
+        # - GitHub has a first_patched_version field,
+        #   but rubysec advisory needs a ruby version spec
+        #
+        # The unnaffected_versions field is similarly not directly available
```
reedloden, line 303: unaffected*

lib/github_advisory_sync.rb
SkipKayhil: I figured it out, `YAML.dump` is trying to preserve the trailing whitespace in some of the files, causing the weird output seen above.

I opened another PR to remove all of the trailing whitespace from the yaml files: #481
Maybe a linting Github Action could be added to prevent any occurrences in the future?

I'll update this PR to use `load`/`dump` tomorrow

lib/github_advisory_sync.rb
```diff
@@ -199,110 +208,120 @@ def ghsa_id
...
+
+      advisory["cvss"]["score"].to_f
     end
 
     def external_reference
-      github_advisory_graphql_object["references"].find do |ref|
-        next if ref["url"].start_with?("https://nvd.nist.gov/vuln/detail/")
-        ref["url"]
+      advisory["references"].find do |ref|
+        !ref["url"].start_with?("https://nvd.nist.gov/vuln/detail/")
```
reedloden, line 236: Eventually, would love to pick the first one as the main URL and add the others as related URLs at the bottom. https://github.com/rubysec/ruby-advisory-db/issues/402

reedloden (commented):
Looking good. Looks like you have a few more things you want to change, and I have some random comments. But overall, excellent changes. Thanks again!

lib/github_advisory_sync.rb
SkipKayhil: Updated to allow for multiple package names

lib/github_advisory_sync.rb
SkipKayhil: Updated to add any missing information from Github

SkipKayhil:
Thanks for all the feedback! I've updated the PR to handle multiple package names in an advisory (in case that comes up), and it will now update any field from Github that is missing in the file.

After running the script, there are 167 files updated (I'll make another PR for that after this one is resolved). I've attached an example diff from `gems/actionpack/CVE-2015-7576.yml`. The only content change I see is the addition of the `ghsa`, and the rest is key ordering or formatting differences caused by `YAML.dump`. I tried changing the `indentation` option of `YAML.dump` but it did not change the indentation of arrays (`patched`/`unpatched`)

```diff
diff --git a/gems/actionpack/CVE-2015-7576.yml b/gems/actionpack/CVE-2015-7576.yml
index 300ddec..b5dbdfa 100644
--- a/gems/actionpack/CVE-2015-7576.yml
+++ b/gems/actionpack/CVE-2015-7576.yml
@@ -2,118 +2,115 @@
 gem: actionpack
 framework: rails
 cve: 2015-7576
-date: 2016-01-25
-url: "https://groups.google.com/forum/#!topic/rubyonrails-security/ANv0HDHEC3k"
-
+ghsa: p692-7mm3-3fxg
+url: https://groups.google.com/forum/#!topic/rubyonrails-security/ANv0HDHEC3k
 title: Timing attack vulnerability in basic authentication in Action Controller.
-
+date: 2016-01-25
 description: |
-    There is a timing attack vulnerability in the basic authentication support
-    in Action Controller. This vulnerability has been assigned the CVE
-    identifier CVE-2015-7576.
+  There is a timing attack vulnerability in the basic authentication support
+  in Action Controller. This vulnerability has been assigned the CVE
+  identifier CVE-2015-7576.

-    Versions Affected:  All.
-    Not affected:       None.
-    Fixed Versions:     5.0.0.beta1.1, 4.2.5.1, 4.1.14.1, 3.2.22.1
+  Versions Affected:  All.
+  Not affected:       None.
+  Fixed Versions:     5.0.0.beta1.1, 4.2.5.1, 4.1.14.1, 3.2.22.1

-    Impact
-    ------
-    Due to the way that Action Controller compares user names and passwords in
-    basic authentication authorization code, it is possible for an attacker to
-    analyze the time taken by a response and intuit the password.
+  Impact
+  ------
+  Due to the way that Action Controller compares user names and passwords in
+  basic authentication authorization code, it is possible for an attacker to
+  analyze the time taken by a response and intuit the password.

-    For example, this string comparison:
+  For example, this string comparison:

-      "foo" == "bar"
+    "foo" == "bar"

-    is possibly faster than this comparison:
+  is possibly faster than this comparison:

-      "foo" == "fo1"
+    "foo" == "fo1"

-    Attackers can use this information to attempt to guess the username and
-    password used in the basic authentication system.
+  Attackers can use this information to attempt to guess the username and
+  password used in the basic authentication system.

-    You can tell you application is vulnerable to this attack by looking for
-    `http_basic_authenticate_with` method calls in your application.
+  You can tell you application is vulnerable to this attack by looking for
+  `http_basic_authenticate_with` method calls in your application.

-    All users running an affected release should either upgrade or use one of
-    the workarounds immediately.
+  All users running an affected release should either upgrade or use one of
+  the workarounds immediately.

-    Releases
-    --------
-    The FIXED releases are available at the normal locations.
+  Releases
+  --------
+  The FIXED releases are available at the normal locations.

-    Workarounds
-    -----------
-    If you can't upgrade, please use the following monkey patch in an initializer
-    that is loaded before your application:
+  Workarounds
+  -----------
+  If you can't upgrade, please use the following monkey patch in an initializer
+  that is loaded before your application:

-    ```
-    $ cat config/initializers/basic_auth_fix.rb
-    module ActiveSupport
-      module SecurityUtils
-        def secure_compare(a, b)
-          return false unless a.bytesize == b.bytesize
+  ```
+  $ cat config/initializers/basic_auth_fix.rb
+  module ActiveSupport
+    module SecurityUtils
+      def secure_compare(a, b)
+        return false unless a.bytesize == b.bytesize

-          l = a.unpack "C#{a.bytesize}"
+        l = a.unpack "C#{a.bytesize}"

-          res = 0
-          b.each_byte { |byte| res |= byte ^ l.shift }
-          res == 0
-        end
-        module_function :secure_compare
+        res = 0
+        b.each_byte { |byte| res |= byte ^ l.shift }
+        res == 0
+      end
+      module_function :secure_compare

-        def variable_size_secure_compare(a, b)
-          secure_compare(::Digest::SHA256.hexdigest(a), ::Digest::SHA256.hexdigest(b))
-        end
-        module_function :variable_size_secure_compare
+      def variable_size_secure_compare(a, b)
+        secure_compare(::Digest::SHA256.hexdigest(a), ::Digest::SHA256.hexdigest(b))
       end
+      module_function :variable_size_secure_compare
     end
-
-    module ActionController
-      class Base
-        def self.http_basic_authenticate_with(options = {})
-          before_action(options.except(:name, :password, :realm)) do
-            authenticate_or_request_with_http_basic(options[:realm] || "Application") do |name, password|
-              # This comparison uses & so that it doesn't short circuit and
-              # uses `variable_size_secure_compare` so that length information
-              # isn't leaked.
-              ActiveSupport::SecurityUtils.variable_size_secure_compare(name, options[:name]) &
-                ActiveSupport::SecurityUtils.variable_size_secure_compare(password, options[:password])
-            end
+  end
+
+  module ActionController
+    class Base
+      def self.http_basic_authenticate_with(options = {})
+        before_action(options.except(:name, :password, :realm)) do
+          authenticate_or_request_with_http_basic(options[:realm] || "Application") do |name, password|
+            # This comparison uses & so that it doesn't short circuit and
+            # uses `variable_size_secure_compare` so that length information
+            # isn't leaked.
+            ActiveSupport::SecurityUtils.variable_size_secure_compare(name, options[:name]) &
+              ActiveSupport::SecurityUtils.variable_size_secure_compare(password, options[:password])
           end
         end
       end
     end
-    ```
+  end
+  ```


-    Patches
-    -------
-    To aid users who aren't able to upgrade immediately we have provided patches for
-    the two supported release series. They are in git-am format and consist of a
-    single changeset.
+  Patches
+  -------
+  To aid users who aren't able to upgrade immediately we have provided patches for
+  the two supported release series. They are in git-am format and consist of a
+  single changeset.

-    * 4-1-basic_auth.patch - Patch for 4.1 series
-    * 4-2-basic_auth.patch - Patch for 4.2 series
-    * 5-0-basic_auth.patch - Patch for 5.0 series
+  * 4-1-basic_auth.patch - Patch for 4.1 series
+  * 4-2-basic_auth.patch - Patch for 4.2 series
+  * 5-0-basic_auth.patch - Patch for 5.0 series

-    Please note that only the 4.1.x and 4.2.x series are supported at present. Users
-    of earlier unsupported releases are advised to upgrade as soon as possible as we
-    cannot guarantee the continued availability of security fixes for unsupported
-    releases.
+  Please note that only the 4.1.x and 4.2.x series are supported at present. Users
+  of earlier unsupported releases are advised to upgrade as soon as possible as we
+  cannot guarantee the continued availability of security fixes for unsupported
+  releases.

-    Credits
-    -------
-
-    Thank you to Daniel Waterworth for reporting the problem and working with us to
-    fix it.
+  Credits
+  -------

+  Thank you to Daniel Waterworth for reporting the problem and working with us to
+  fix it.
 cvss_v2: 4.3
 cvss_v3: 3.7
-
 patched_versions:
-  - ">= 5.0.0.beta1.1"
-  - "~> 4.2.5, >= 4.2.5.1"
-  - "~> 4.1.14, >= 4.1.14.1"
-  - "~> 3.2.22.1"
+- ">= 5.0.0.beta1.1"
+- "~> 4.2.5, >= 4.2.5.1"
+- "~> 4.1.14, >= 4.1.14.1"
+- "~> 3.2.22.1"
```

lib/github_advisory_sync.rb
phillmv: @reedloden the schema changes enabling that shipped the same time we made this change https://github.com/rubysec/ruby-advisory-db/pull/479#discussion_r655432437

i think in practice it's just extremely rare 🤔 might not have happened yet in ruby land, i'd have to poke the API to find out :P.

phillmv:
My eyes are too tired to give this justice right now, I'll circle back on this tomorrow. @SkipKayhil if the `YAML.load(old) == YAML.load(new).except('ghsa')` matches, that's good enough for me.

phillmv (approved):
This seems fine to me! Thanks for the update 😄. Ideally we'd have tests 🤔 but if it works that's good enough for now.

PR #481, opened 2021-06-22
remove trailing whitespaces from yml files

SkipKayhil:
Using command: `sed -i 's/[ \t]*$//' gems/**/*.yml`

(requires gsed)

Trailing whitespace in multiline text blocks prevents them from being
dumped in the 'pretty', `|` format with YAML.dump

Ref: #479


reedloden:
Can we add a linting rule to detect and block on this in the future?

reedloden (approved):

PR #482, opened 2021-06-22
add Github Super-Linter action

SkipKayhil:
Related: 8c1cd7dc, #481

Ensure that lint errors like trailing whitespace are not added going
forward to prevent errors with consistent parse/dump of YAML files.

In its current state it will run all of the linters listed [here](https://github.com/github/super-linter#supported-linters) only on files added/changed in a PR

If we just want it to run yamllint (or rubocop, etc.) then we can specify which linters to run

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,55 @@
+---
+#################################
+#################################
+## Super Linter GitHub Actions ##
+#################################
+#################################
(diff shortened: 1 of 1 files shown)


reedloden:
I ended up adding `yamllint` separately for now... This super linter looks cool, so may revisit that at some point. Thanks!

PR #484, opened 2021-06-26
add missing ghsa to vulnerabilities using sync

SkipKayhil:
The sync script will now update vulnerabilities with any missing information
from the Github advisory database, so this is the result of running it.

There are a lot of changes, so I verified the diff with this ruby
script:

```ruby
require 'yaml'

BRANCH = "update-ghsa"
files_to_check = `git diff-tree --no-commit-id --name-only -r #{BRANCH}`.split

old_data = {}
new_data = {}

system("git switch master")

files_to_check.each do |file|
  old_data[file] = YAML.load_file(file)
end

system("git switch #{BRANCH}")

files_to_check.each do |file|
  new_data[file] = YAML.load_file(file)
end

files_to_check.each do |file|
  next if old_data[file] == new_data[file].except('ghsa')

  puts '====='
  puts file
  puts 'NEW KEYS'
  p new_data[file].to_a - old_data[file].to_a
  puts 'REMOVED KEYS'
  p old_data[file].to_a - new_data[file].to_a
end
```

There are four changes other than ghsa additions:
- gems/actionpack/CVE-2021-22902.yml, gems/actionpack/CVE-2021-22903.yml
and gems/actionpack/CVE-2021-22904.yml also have cvss_v3 scores added
- gems/rest-client/CVE-2015-1820.yml had a blank cvss_v2 score, so the
  key was removed in the sync


SkipKayhil:
The verification script also found a few keys that are not listed in the schema in the README:

https://github.com/rubysec/ruby-advisory-db/blob/b562149c00ae79cf04b892a63c88f34ca07d7159/gems/moped/CVE-2015-4410.yml#L11-L13

https://github.com/rubysec/ruby-advisory-db/blob/b562149c00ae79cf04b892a63c88f34ca07d7159/gems/rubygems-update/CVE-2019-8325.yml#L3

I didn't include changes to those files because I wasn't sure if those keys should be kept. If we want to keep these, I can add them to the whitelisted fields in the sync script, so it won't try to remove them in the future

gems/actionview/CVE-2016-0752.yml
```diff
@@ -84,12 +84,9 @@ description: |
   Credits
   -------
   Thanks John Poulin for reporting this!
-
 cvss_v2: 5.0
 cvss_v3: 7.5
-
-# "~> 3.2.22.1" is found in gems/actionpack/CVE-2016-0752.yml
```
SkipKayhil, line 91: It looks like `load/dump` removes comments. Should these be converted to a `comments` field or something similar so they are kept around?

reedloden (approved):
Thanks for the great work! I pulled out a lot of the one-off things you mentioned into separate commits and addressed. I rebased your branch and re-ran the script. Things look good, so time to merge.

PR #494, opened 2021-12-14
Sync with GitHub Security Advisories

radar:
* Add CVE-2021-44528 for actionpack (announced today)
* Add CVE-2018-14042 for bootstrap
* Add CVE-2017-18258 for Nokogiri

### DIFF
--- a/gems/actionpack/CVE-2021-44528.yml
+++ b/gems/actionpack/CVE-2021-44528.yml
@@ -1,0 +1,117 @@
+---
+gem: actionpack
+cve: 2021-44528
+ghsa: qphc-hf5q-v8fc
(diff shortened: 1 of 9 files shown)


phillmv:
Not at my computer or else i'd edit it myself but do note there's some remaining fields to edit from the sync :). I can circle back later tonight.

reedloden:
Unfortunately, there's a lot of manual work that has to be done as part of any GHSA import, so you can't just run the script and commit it, sadly.

Thank you for the poke. I went ahead and added the one missing CVE in https://github.com/rubysec/ruby-advisory-db/commit/56df670d719ad2d7ab214c97aca927875cb4083f.

Everything else included here is already included in other IDs as a `related` item. GHSA's database duplicates several things that we combine all together for our purposes.

As such, I'm going to go ahead and close this. Thanks for helping the RubySec community!

PR #498, opened 2022-03-04
Add unaffected versions for CVE-2022-24722

IKrizic:
### **Problem:**
Current version of view_component gem we are using on a project is **2.26.1** and checking our Gemfile.lock file against ruby-advisory database by using [bundler-audit](https://github.com/rubysec/bundler-audit) reports CVE-2022-24722.
According to GitHub Advisory view_component versions affected by CVE-2022-24722 are:
```
>= 2.31.0, < 2.31.2
>= 2.32.0, < 2.49.1
```
###  **Solution:**
  Add “< 2.31.0” to  unaffected_versions in [CVE-2022-24722.yml](https://github.com/rubysec/ruby-advisory-db/commit/4a57fd51236cbb0b03d13583f1bf75d9f3d0ced0#diff-655f132a9ea2f6e77397adf43790938ebbaea351027a0f728ff019c5f7649962)

### DIFF
--- a/gems/view_component/CVE-2022-24722.yml
+++ b/gems/view_component/CVE-2022-24722.yml
@@ -15,6 +15,8 @@ description: |
 
   Avoid passing user input to the `translate` function, or sanitize the inputs
   before passing them.
+unaffected_versions:
+- "< 2.31.0"
 patched_versions:
 - ">= 2.49.1"
 - "~> 2.31.2"


postmodern:
Those version ranges confused me at first, but I ended up implementing the same `unaffected_versions` in 3ed3fc8794da3a4492a045efa175b137044c11a3. Does that fix things on your end?

reedloden:
Have we validated that by looking at the code? I've found GHSA version strings to sometimes be unreliable.

It also doesn't help that the advisory incorrect states:
> Versions 2.29.1 and 2.49.1 have been released and fully mitigate the vulnerability.

... when it's actually 2.31.2 and 2.49.1

reedloden:
Looks like the `ViewComponent::Translatable` class was added in v2.29.0 (https://github.com/github/view_component/blob/main/docs/CHANGELOG.md#2290).

However, v2.31.0 did significantly overhaul it, likely adding the vulnerable code. So, `<2.31.0` is probably correct indeed.

reedloden:
@IKrizic I'm going to close this out, but let us know if this doesn't fix it for some reason.

IKrizic:
@postmodern I feel ya, I was confused as well. 
I agree with @reedloden's conclusions and everything is ok now on our side so I think we are good 👍 

Thanks guys, that was fast! ⚡

PR #500, opened 2022-04-14
Add latest ruby CVEs

ylecuyer:
Those 2 new CVEs are impacting ruby directly and are not in the github registry of CVEs (hence are not synced here):

* https://www.ruby-lang.org/en/news/2022/04/12/buffer-overrun-in-string-to-float-cve-2022-28739/
* https://www.ruby-lang.org/en/news/2022/04/12/double-free-in-regexp-compilation-cve-2022-28738/


reedloden:
Thanks, @ylecuyer!

PR #511, opened 2022-07-26
Add patched version for CVE-2021-41816 and CVE-2021-41819 on cgi

yano3:
Add patched version for CVE-2021-41816 and CVE-2021-41819 on cgi. These are fixed in cgi 0.1.0.1 bundled with Ruby 2.7.5 or later.

- ref: https://github.com/ruby/ruby/commit/f69aeb83146be640995753667fdd6c6f157527f5

### DIFF
--- a/gems/cgi/CVE-2021-41816.yml
+++ b/gems/cgi/CVE-2021-41816.yml
@@ -18,6 +18,7 @@ description: |-
   2.6 is not vulnerable.
 cvss_v3: 7.5
 patched_versions:
+- "~> 0.1.0.1"
 - "~> 0.1.1"
 - "~> 0.2.1"
 - ">= 0.3.1"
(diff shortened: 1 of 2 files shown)


postmodern:
I assume `cgi-0.1.0.1` will be released to [rubygems.org](https://rubygems.org/gems/cgi/versions) or ruby-core will update the `cgi.gemspec` file in `lib/ruby/gems/2.7.0/specifications/default/`? That way `bundler` and `bundle-audit` will detect the `cgi`' gem's new updated version.

yano3:
`cgi-0.1.0.1` is already included in Ruby `2.7.5` and `2.7.6`.

```
$ docker run -it --rm --name ruby ruby:2.7.5 ls -l /usr/local/lib/ruby/gems/2.7.0/specifications/default/ | grep cgi
-rw-r--r-- 1 root root  1053 Mar 30 23:05 cgi-0.1.0.1.gemspec
```

I my use case, [Trivy](https://github.com/aquasecurity/trivy) that uses ruby-advisory-db as data source [will have false positives](https://gist.github.com/yano3/78cb494fa4783ba2a2bd82b18d2777c9) even though they have been fixed.

PR #514, opened 2022-07-27
docs: fix broken link for OSVDB

lirantal:
README link was broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,12 +122,12 @@ GH_API_TOKEN=<your GitHub API Token> bundle exec rake sync_github_advisories
 
 Please see [CONTRIBUTORS.md].
 
-This database also includes data from the [Google's Open Source Vulnerability Database][OSVDB]
+This database also includes data from the [Open Sourced Vulnerability Database][OSVDB]
 developed by the Open Security Foundation (OSF) and its contributors.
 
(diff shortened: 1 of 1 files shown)


reedloden (changes requested):
OSVDB is correct here. Perhaps we should update the URL to be https://en.wikipedia.org/wiki/Open_Source_Vulnerability_Database instead.

lirantal:
Is that still relevant though? I was under the impression that this is part of the vulnerabilities sourced into ruby-advisory-db but if that doesn't exist anymore should we just remove it entirely?

postmodern:
Google's OSV.dev is entirely different from the old OSVDB (where Open Source is meant to refer to Open Source Intelligence, not Open Source Software).

phillmv:
> Is that still relevant though? I was under the impression that this is part of the vulnerabilities sourced into ruby-advisory-db but if that doesn't exist anymore should we just remove it entirely?

Looks like we still use `osvdb` ids [for really old vulns](https://github.com/rubysec/ruby-advisory-db/blob/9f906cf530f23986f07cfee5230b3f2e5779aedb/gems/as/OSVDB-112683.yml#L3)

I second Reed's suggestion for pointing the URL to the wiki page 😃.

reedloden (approved):

PR #515, opened 2022-09-15
Jruby CVE-2022-25857

ylecuyer:
Hello, there is a CVE for jruby, I'm not 100% sure of how I did the report as it is a transitive dependency but in the end it gets bundled into the jruby which is available for download.


postmodern:
@ylecuyer you accidentally committed the `.swp` file. `git rm` and `git commit --amend`.

ylecuyer:
Ouch indeed I fixed that :+1:

PR #531, opened 2022-12-13
add CVEs for loofah and rails-html-sanitizer

flavorjones:
This morning's GHSAs and CVEs for loofah and rails-html-sanitizer.


reedloden (approved):
Generally, we don't include credits and references should be converted to use `related`. But we can fix those things after the fact easily.

flavorjones:
Ah, sorry about that. I'll submit a PR later tonight to fix. Thank you!

PR #532, opened 2022-12-13
fix: remove credits and use `related` for references

flavorjones:
Per comments at #531, remove credits and use `related` for references. Also remove redundant CVE information from descriptions.


reedloden (approved):
Thank you so much!

PR #538, opened 2023-01-23
Consider Rails LTS 5.2.8.15

tkdn:
Rails LTS should be included.

- https://makandracards.com/railslts/508019-rails-5-2-lts-changelog#section-jul-21st-2022-version-5-2-8-15

### DIFF
--- a/gems/actionpack/CVE-2023-22792.yml
+++ b/gems/actionpack/CVE-2023-22792.yml
@@ -12,7 +12,7 @@ description: |
 
   Versions Affected: >= 3.0.0
   Not affected: < 3.0.0
-  Fixed Versions: 6.1.7.1, 7.0.4.1
+  Fixed Versions: 5.2.8.15, 6.1.7.1, 7.0.4.1
 
   # Impact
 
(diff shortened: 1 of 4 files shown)


postmodern:
This might confuse people who are not familiar with Rails LTS, since Rails LTS has released versions which stock Rails has not. Could you add `(rails LTS)` after the `5.x` versions in the description text, and a `# rails LTS` commend after the `~> 5.x, >= 5.x.y.z` lines?

tkdn:
@postmodern 
Thanks for the review. I've added comments for rails (LTS). 50d6e1d

reedloden (approved):
Thanks for submitting this!

rea-jonpad:
> This might confuse people who are not familiar with Rails LTS, since Rails LTS has released versions which stock Rails has not. Could you add `(rails LTS)` after the `5.x` versions in the description text, and a `# rails LTS` commend after the `~> 5.x, >= 5.x.y.z` lines?

Because of the way downstream systems such as bundler-audit use the YAML, the `# rails LTS` comment appended after the `~> 5.x, >= 5.x.y.z` lines is effectively ignored by the YAML parser.

Then bundler-audit recommends upgrading to the Rails LTS gem, which is a commercial offering, and not available without a [license, username & password](https://makandracards.com/railslts/508020-installing-rails-5-2-lts).

I realize that in some ways, this is bundler-audit's problem. But with the "# (Rails LTS)" effectively being stripped out by the YAML parser, bundler-audit has no way to identify such gems as being restricted, and not widely available, however it still recommends it as a fix.

Perhaps fixed versions from external third parties could have additional metadata and be marked as such?

PR #570, opened 2023-04-10
Modified github_advisory_sync.rb's new_data variable to add 'patched_versions' and related:'s url…

jasnow:
Modified github_advisory_sync.rb's **new_data** variable to add the "patched_versions" and "related:'"/"url:" sections.

The new (unreleased 1-line shell) post processor does the rest to get GHSA files ready for ruby-advisory-db repo. Long term this shell will be turned into Ruby code.

### DIFF
--- a/lib/github_advisory_sync.rb
+++ b/lib/github_advisory_sync.rb
@@ -351,9 +351,9 @@ module GitHub
       File.open(filename_to_write, "w") do |file|
         # create an automatically generated advisory yaml file
         file.write new_data.merge(
-          { "patched_versions" => vulnerabilities },
-          { "related" => { "url"  => advisory["references"] } }
-        ).to_yaml
+          { "patched_versions" => vulnerabilities,
+            "related" => { "url"  => advisory["references"] }
+          } ).to_yaml
 
         # The data we just wrote is incomplete,
         # and therefore should not be committed as is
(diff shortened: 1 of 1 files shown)


jasnow:
@reedloden - FYI

lib/github_advisory_sync.rb
```diff
@@ -350,7 +350,10 @@ def create(package)
       Dir.mkdir dir_to_write unless Dir.exist?(dir_to_write)
       File.open(filename_to_write, "w") do |file|
         # create an automatically generated advisory yaml file
-        file.write new_data.to_yaml
+        file.write new_data.merge(
+          { "patched_versions" => vulnerabilities },
+          { "related:\nurl:"  => advisory["references"] }
```
postmodern, line 355: I think that should be `{"related" => {"url" => ...}}`.

postmodern, line 355: Also these two hashes could probably be combined.

postmodern (changes requested):
Use nested Hashes instead of using `\n` in Hash keys (I'm pretty sure YAML will escape those). Also the Hashes can probably be merged into one `new_data.merge(...)` Hash.

lib/github_advisory_sync.rb
jasnow: fuxed

jasnow: fixed

jasnow:
Thanks for the clarification

PR #575, opened 2023-04-11
[GHSA SYNC2] Adding four advisories based on CVE-2014-3248 advisory

jasnow:
Adding four advisories (files) based on CVE-2014-3248 advisory.

* These files were imported using "lib/github_advisory_sync.rb" script.
* Only edit was making 2-line title into 1.
* Checked "yamllint" (clean) and ran "rake" (clean).

### DIFF
--- a/gems/bootstrap/CVE-2018-14042.yml
+++ b/gems/bootstrap/CVE-2018-14042.yml
@@ -2,7 +2,7 @@
 gem: bootstrap
 cve: 2018-14042
 ghsa: 7mvr-5x2g-wfc8
-url: https://github.com/twbs/bootstrap/issues/26423
+url: https://github.com/twbs/bootstrap/issues/26428
 title: Bootstrap Cross-site Scripting vulnerability
 date: 2018-09-13
 description:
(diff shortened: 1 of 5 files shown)


gems/bootstrap/CVE-2018-14042.yml
```diff
@@ -0,0 +1,33 @@
+---
+gem: bootstrap
+cve: 2018-14042
+ghsa: 7mvr-5x2g-wfc8
+url: https://github.com/twbs/bootstrap/issues/26423
```
postmodern: I think the primary URL for this advisory should be `https://github.com/twbs/bootstrap/issues/26628`, which explains the vulnerability.

postmodern (changes requested):
Primary URL for `gems/bootstrap/CVE-2018-14042.yml` should point to the actual advisory/vulnerability issue.

gems/bootstrap/CVE-2018-14042.yml
jasnow: fixed

Issue #573, opened 2023-04-11
Convert shell-based post processor code + manual steps into "lib/github_advisory_sync.rb" code


jasnow:
# Convert shell-based post processor code + manual steps into "lib/github_advisory_sync.rb" (Ruby) code
## Currently fixed by post processor run
* [ ] 1. Remove code from "GitHub advisory data below" to end of file. Also remove 2 blank lines before "GitHub advisory data below".
* [ ] 2. Remove "url:" from "related:"/"url:" (reference) urls.
* [ ] 3. Add description next line feature ("description: |\n  ")
## Currently manual or removed after post processor run
* [ ] 4. Complete #571 - GHSA's: package field to "patched_versions" field
* [ ] 5. Remove or add issue to automate lines containing "unaffected_versions:" and "OPTIONAL: FILL IN SEE BELOW".
* [ ] 6. Remove or add issue to automate line containing [cvss_v3: "\<FILL IN IF AVAILABLE\>"] line.

jasnow:
Pruning

Issue #571, opened 2023-04-10
[GHSA SYNC2] Change "lib/github_advisory_sync.rb" (ruby) script to convert GHSA's: package field to "patched_versions" field


jasnow:
Change "lib/github_advisory_sync.rb" (ruby) script to convert GHSA's: json package field
```
- package:
    name: rails_admin
    ecosystem: RUBYGEMS
  vulnerableVersionRange: "< 1.3.0"
  firstPatchedVersion:
    identifier: 1.3.0
```
To ruby-advisory-db's patched_versions yaml field. [Single Patch Example](https://github.com/rubysec/ruby-advisory-db/blob/master/gems/rails_admin/CVE-2017-12098.yml)
```
patched_versions:
  - ">= 1.3.0"
```
Remember the use case with multiple patched versions. [Mutliple Patches Example](https://github.com/rubysec/ruby-advisory-db/blob/master/gems/uri/CVE-2023-28755.yml)
```
patched_versions:
  - "~> 0.10.0.1"
  - "~> 0.10.2"
  - "~> 0.11.1"
  - ">= 0.12.1"
```
4/14/2023: UPDATE: @postmodern's PR 577 feedback regarding patched_version ranges
 * Did you test your changes with a GHSA advisory that had multiple patched_versions ranges? Such as GHSA-6x85-j5j2-27jx.
   * When a patched version range has an upper and lower bound (ex: >= 2.1.0 and < 3.0.0) then it needs to be represented as a comma separated YAML String (ex: >= 2.1.0, < 3.0.0) which bundler-audit can
parse. I'm not sure if GitHub's GraphQL end-point returns patched_versions as an Array of Strings
(ex: [">= 2.1.0, < 3.0.0", ...]) or possibly an Array of Arrays of Strings (ex: [[">= 2.1.0", "< 3.0.0"], ...])?

jasnow:
Pruning

PR #577, opened 2023-04-13
#573: Changed "lib/github_advisory_sync.rb" to remove the need for a shell-based post processor

jasnow:
## #573: Changed the ruby code of "lib/github_advisory_sync.rb" to remove the need for a shell-based post processor.

 * Issue 573 punch list items 2 and 3 required code changes.
 * Issue 573 punch list items 1, 5, and 6 were done by commenting out the code instead of deleting it.
 * Issue 573 punch list item 4 will be tracked in #571.
 * Therefore #573 will be considered fixed when PR is accepted.
    * Limitation: Must start "patched_versions" 2nd line in column 1 for "yamllint" to be happy.

### DIFF
--- a/lib/github_advisory_sync.rb
+++ b/lib/github_advisory_sync.rb
@@ -308,7 +308,7 @@ module GitHub
         "ghsa"        => ghsa_id[5..],
         "url"         => external_reference,
         "title"       => advisory["summary"],
-        "description" => advisory["description"],
+        "description" => advisory["description"].gsub(/description: (.*)/, 'description: |\n\1'),
         "cvss_v3"     => cvss,
       }.compact
     end
@@ -342,9 +342,10 @@ module GitHub
       filename_to_write = package.filename
 
       new_data = package.merge_data(
-        "cvss_v3"             => ("<FILL IN IF AVAILABLE>" unless cvss),
-        "patched_versions"    => ["<FILL IN SEE BELOW>"],
-        "unaffected_versions" => ["<OPTIONAL: FILL IN SEE BELOW>"]
+        "cvss_v3"             => (nil unless cvss),
+#HID        "cvss_v3"             => ("<FILL IN IF AVAILABLE>" unless cvss),
+#HID        "patched_versions"    => ["<FILL IN SEE BELOW>"],
+#HID        "unaffected_versions" => ["<OPTIONAL: FILL IN SEE BELOW>"]
       )
 
       FileUtils.mkdir_p(File.dirname(filename_to_write))
@@ -353,7 +354,7 @@ module GitHub
         file.write new_data.merge(
           "patched_versions" => vulnerabilities,
           "related" => {
-            "url"  => advisory["references"]
+            "url"  => advisory["references"].map! { |url| url['url'] }
           }
         ).to_yaml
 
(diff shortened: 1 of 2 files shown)


jasnow:
Note that the sample advisory file is 2 days old.

lib/github_advisory_sync.rb
```diff
@@ -308,7 +308,7 @@ def to_h
         "ghsa"        => ghsa_id[5..],
         "url"         => external_reference,
         "title"       => advisory["summary"],
-        "description" => advisory["description"],
+        "description" => advisory["description"].gsub(/description: (.*)/, 'description: |\n\1'),
```
postmodern, line 311: That `gsub` call is probably not going to do anything. The advisory metadata is pulled via GitHub's GraphQL end-point via `retrieve_all_rubygem_vulnerabilities`, so it won't contain the YAML `description: ...` key, it will just contain the description text.

lib/github_advisory_sync.rb
```diff
@@ -342,9 +342,10 @@ def create(package)
       filename_to_write = package.filename
 
       new_data = package.merge_data(
-        "cvss_v3"             => ("<FILL IN IF AVAILABLE>" unless cvss),
-        "patched_versions"    => ["<FILL IN SEE BELOW>"],
-        "unaffected_versions" => ["<OPTIONAL: FILL IN SEE BELOW>"]
+        "cvss_v3"             => (nil unless cvss),
+#HID        "cvss_v3"             => ("<FILL IN IF AVAILABLE>" unless cvss),
+#HID        "patched_versions"    => ["<FILL IN SEE BELOW>"],
+#HID        "unaffected_versions" => ["<OPTIONAL: FILL IN SEE BELOW>"]
```
postmodern, line 348: Did you test your changes with a GHSA advisory that had multiple `patched_versions` ranges? Such as [GHSA-6x85-j5j2-27jx](https://github.com/advisories/GHSA-6x85-j5j2-27jx). When a patched version range has an upper and lower bound (ex: `>= 2.1.0 and < 3.0.0`) then it needs to be represented as a comma separated YAML String (ex: `>= 2.1.0, < 3.0.0`) which `bundler-audit` can parse. I'm not sure if GitHub's GraphQL end-point returns `patched_versions` as an Array of Strings (ex: `[">= 2.1.0, < 3.0.0", ...]`) or possibly an Array of Arrays of Strings (ex: `[[">= 2.1.0", "< 3.0.0"], ...]`)?

If the new code changes work perfectly, you can simply delete the commented out code, as it's no longer needed.

lib/github_advisory_sync.rb
```diff
@@ -353,7 +354,7 @@ def create(package)
         file.write new_data.merge(
           "patched_versions" => vulnerabilities,
           "related" => {
-            "url"  => advisory["references"]
+            "url"  => advisory["references"].map! { |url| url['url'] }
```
postmodern, line 357: You probably want `.map` which returns a new `Array`, instead of modifying in-place which changes the contents of `advisory["references"]` permanently.

lib/github_advisory_sync.rb
```diff
@@ -342,9 +342,10 @@ def create(package)
       filename_to_write = package.filename
 
       new_data = package.merge_data(
-        "cvss_v3"             => ("<FILL IN IF AVAILABLE>" unless cvss),
-        "patched_versions"    => ["<FILL IN SEE BELOW>"],
-        "unaffected_versions" => ["<OPTIONAL: FILL IN SEE BELOW>"]
+        "cvss_v3"             => (nil unless cvss),
```
postmodern, line 345: `(nil unless cvss)` could be simplified to just `cvss`.

lib/github_advisory_sync.rb
```diff
@@ -373,9 +374,9 @@ def create(package)
         # The second block of yaml in a .yaml file is ignored (after the second "---" line)
         # This effectively makes this data a large comment
         # Still it should be removed before the data goes into rubysec
-        file.write "\n\n# GitHub advisory data below - **Remove this data before committing**\n"
-        file.write "# Use this data to write patched_versions (and potentially unaffected_versions) above\n"
-        file.write advisory.merge("vulnerabilities" => vulnerabilities).to_yaml
+#HID        file.write "\n\n# GitHub advisory data below - **Remove this data before committing**\n"
+#HID        file.write "# Use this data to write patched_versions (and potentially unaffected_versions) above\n"
+#HID        file.write advisory.merge("vulnerabilities" => vulnerabilities).to_yaml
```
postmodern, line 379: RE commented out code, ditto here. If it's no longer needed remove it.

postmodern (changes requested):
I had some questions about how `patched_versions` and `unpatched_versions` are being formatted. I also spotted some redundant code.

jasnow:
I read and triaged your 6 comments.
 * Your "multiple patched_version" text will be moved to issue 571 since I am manually do that field currently.
 * With most of your comments requiring coding changes, this PR will be canceled.

PR #587, opened 2023-04-28
Two GHSA-based mruby advisories: One with huntr.dev data; one without

jasnow:
Two GHSA-based mruby advisories.
* One with huntr.dev data; one without  huntr.dev data.
* 37 more mruby advisories to go after these two.


postmodern:
LGTM Thanks!

PR #588, opened 2023-04-30
Adding the remaining 18 GHSA-based mruby huntr.dev advisories

jasnow:
Adding the remaining 18 GHSA-based mruby huntr.dev advisories.
They are all similar (between 19-22 lines per file).
yamllint and "rake" are green.


postmodern:
Thanks!

PR #599, opened 2023-05-11
5 new GHSA-based 2012/2013 advisories

jasnow:
5 new GHSA-based 2012/2013 advisories
yamllint and rake (green)


postmodern (approved):
LGTM

PR #603, opened 2023-05-15
Standardized "description:" format

jasnow:
Wrote a script to fix the existing (128) yml files' "description:" field to only use "|":
  * Along with these patterns: ": [a-zA-Z('\"]", I found these:
      5 description:
      2 description: >-
      4 description: |+
     53 description: |-
      2 description: |2
      1 description: |2-

Based on your title: test work, I added this test to spec/advisory_example.rb:
```
      it "must not be one line" do
        is_expected.to include("\n")
      end
```

Ran
```
cat $(find gems rubies -type f) |grep "description:"  |sort |uniq -c
```
and got:
```
    865 description: |
```

Then ran "yamllint" and "rake (both green).


postmodern (approved):
LGTM, but will cherry-pick the last commit to avoid pulling in the other commits from previous PRs that were already squash+merged.

postmodern:
Manually cherry-picked 635a614 into master.

jasnow:
Thanks - got rid of extra commits

PR #604, opened 2023-05-16
Update Gemfile.lock file

jasnow:
Update Gemfile.lock file by running "bundle update". 
Ran "rake" (green).


jasnow:
oh well

postmodern:
I was considering merging this? Although maybe it's better to just remove the `Gemfile.lock` file and set `~> X.Y` version ranges in `Gemfile`. `Gemfile.lock` files are typically only committed for apps that are deployed.

jasnow:
> I was considering merging this? Although maybe it's better to just remove the `Gemfile.lock` file and set `~> X.Y` version ranges in `Gemfile`. `Gemfile.lock` files are typically only committed for apps that are deployed.

You are welcome to do as you wish.

adam12:
> `Gemfile.lock` files are typically only committed for apps that are deployed.

Just FYI, this was the opininion held for the longest time, but the Bundler team changed their mind about it a while back and decided that it was beneficial for libraries as well, but there should be provisions during CI to delete the lock file before installing the bundle (to get the latest libraries that would normally install).

PR #605, opened 2023-05-22
Updated README to match fact that we now support rubies - issue 193

jasnow:
Updated README to match fact that we now support rubies - issue 193
Also removed duplicate text from CONTRIBUTING.md file to ease maintenance.


postmodern:
Let's instead create two separate sections for the `rubies/` advisory schemas vs. the `gems/` advisory schemas. We could also put those in a new `SCHEMA.md` file.

postmodern:
See 183a5806ea381630329c87ec76361089dd9d52ac and 2147579a3b6af5ea1e587e679033e834253649af.

PR #610, opened 2023-05-30
Updated ruby.yml schema to match README file

jasnow:
Found that the ruby.yml schema did not match the updated README file so fixed it.


postmodern (approved):

jasnow:
Thanks

PR #614, opened 2023-06-03
Ran GHSA sync script and got 2 modified and 1 new

jasnow:
Ran GHSA sync script and got 2 modified advisories and 1 new advisories.
Ran yamllint and rake (both green).


postmodern (approved):
LGTM. Not sure why GHSA used the packetstorm link as their primary URL though?

PR #615, opened 2023-06-03
Deleted non-ruby advisory and duplicate OSVDB* advisory

jasnow:
Deleted non-ruby advisory (CVE-2011-3389.yml) and duplicate OSVDB* advisory (mentioned in #613).


postmodern (changes requested):
You seem to have also deleted `rubies/rbx/OSVDB-78119.yml` which seems unrelated to the SSL BEAST advisory. These should be separate commits.

jasnow:
different approach

Issue #238, opened 2016-01-26
Add advisories from SRC:CLR


reedloden:
Just a todo list I figured I should put somewhere more public... Need to add advisories for all these:

ruby_rncryptor / ruby_rncryptor_secured -- https://srcclr.com/security/timing-attacks/ruby/s-1938

spina -- https://srcclr.com/security/cross-site-request-forgery-csrf/ruby/s-1686

logstash-core -- https://srcclr.com/security/factoring-attack-rsa-export-keys-freak/ruby/s-1745
https://srcclr.com/security/man-middle-mitm-attacks/ruby/s-1798

facter -- https://srcclr.com/security/disclosure-amazon-ec2-iam-instance/ruby/s-1508
https://srcclr.com/security/elevation-privileges-untrusted-search/ruby/s-1586

kafo -- https://srcclr.com/security/world-readable-permissions-as-default/ruby/s-740

puppet -- https://srcclr.com/catalog/search#query=type:vulnerability%20puppet

VanessaHenderson:
refinerycms-core 
https://srcclr.com/security/cross-site-scripting-xss-through-title/ruby/s-2168
https://srcclr.com/security/cross-site-scripting-xss-through-alt/ruby/s-2169
https://srcclr.com/security/stored-cross-site-scripting-xss-title/ruby/s-2170
https://srcclr.com/security/cross-site-request-forgery-csrf/ruby/s-2171

VanessaHenderson:
paperclip -- https://srcclr.com/security/denial-service-dos-through-excessive/ruby/s-2242

VanessaHenderson:
admin-upmin, upmin and shoppe -- https://srcclr.com/security/cross-site-request-forgery-csrf-due-to/ruby/s-2266
devise_invitable -- https://srcclr.com/security/cross-site-request-forgery-csrf/ruby/s-2272

reedloden:
activeadmin -- https://srcclr.com/security/cross-site-scripting-xss-through-modal/ruby/s-2276

skorth:
@reedloden regarding spina ruby gem https://github.com/rubysec/ruby-advisory-db/pull/250

jasnow:
Not sure https://srcclr.com is an active web site.

VanessaHenderson:
You are correct, SourceClear was purchased by Veracode. The easiest way to find the links now would be to search the s-* number at https://sca.analysiscenter.veracode.com/vulnerability-database
[...]

PR #616, opened 2023-06-04
Updated 3 and added 3 SRC:CLR advisories with CVE values from issue #238

jasnow:
Updated 3 and added 3 SRC:CLR advisories with CVE values from issue #238.
 * For the existing advisories, I added the updated SRC:CLR URL since SourceClear was purchased by Veracode.
 * For the brand new advisories, I added new files along with the updated SRC:CLR URL.
  * The other 10 URL in #238 do not have a CVE number so they need a CVE before adding them.

### DIFF
--- a/gems/facter/CVE-2015-1426.yml
+++ b/gems/facter/CVE-2015-1426.yml
@@ -1,0 +1,26 @@
+---
+gem: facter
+cve: 2015-1426
+ghsa: j436-h7hm-rx46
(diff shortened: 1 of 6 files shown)


jasnow:
@postmodern - Ready for you.

postmodern (approved):
LGTM

jasnow:
Thanks

PR #618, opened 2023-06-07
OSVDB-78119.yml is duplicate of CVE-2012-5372.yml so deleted it

jasnow:
OSVDB-78119.yml is duplicate of CVE-2012-5372.yml so deleted it.


jasnow:
too much changed

PR #620, opened 2023-06-07
OSVDB-78119.yml is duplicate of CVE-2012-5372.yml so delete it

jasnow:
OSVDB-78119.yml is duplicate of CVE-2012-5372.yml so delete it.


postmodern:
Looks like these are commits from PR #621.

jasnow:
Something went wrong so closing this PR.

PR #621, opened 2023-06-07
Replaced OSVDB-124991.yml with CVE-2015-20108.yml

jasnow:
Replaced OSVDB-124991.yml with CVE-2015-20108.yml.


postmodern (approved):
LGTM

PR #628, opened 2023-06-10
GHSA sync + modifies + few manual edits

jasnow:
GHSA sync + modifies + few manual edits

After running the sync script and removing dups/etc and post-processor script and a few manual edits, the modify summary is:
  * ghsa: field was added
  * Single quotes changed to double quotes per local convention
   * Added double quotes back if removed.
    * Reorder date: field in file.


postmodern (approved):

jasnow:
Thanks

PR #629, opened 2023-06-11
Added 2006 rwiki advisory - CVE-2006-2581.yml

jasnow:
Added 2006 rwiki advisory - CVE-2006-2581.yml.

Very early advisory.

Some advisories in Japanese.

### DIFF
--- a/gems/rwiki/CVE-2006-2581.yml
+++ b/gems/rwiki/CVE-2006-2581.yml
@@ -1,0 +1,39 @@
+---
+gem: rwiki
+cve: 2006-2581
+ghsa: gvhx-gj42-m28v
+url: https://web.archive.org/web/20090501134922/http://www2a.biglobe.ne.jp/~seki/ruby/rwiki.html
+title: RWiki before 2.1.0 has cross-site scripting vulnerability
+date: 2006-05-14
+description: |
+  Cross-site scripting (XSS) vulnerability in Wiki content in
+  RWiki 2.1.0pre1 through 2.1.0 allows remote attackers to inject
+  arbitrary web script or HTML via unknown attack vectors.
+cvss_v2: 4.3
+unaffected_versions:
+  - "< 2.1.0pre1"
+patched_versions:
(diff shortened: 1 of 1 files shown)


gems/rwiki/CVE-2006-2581.yml
```diff
@@ -0,0 +1,39 @@
...
+date: 2006-05-14
+description: |
+  Cross-site scripting (XSS) vulnerability in Wiki content in
+  RWiki 2.1.0pre1 through 2.1.0 allows remote attackers to inject
+  arbitrary web script or HTML via unknown attack vectors.
+cvss_v2: 4.3
+unaffected_versions:
+  - "< 2.1.0pre1"
+patched_versions:
+  - ">= 2.1.0"
```
postmodern, line 16: Google translation of the URL says that `2.1.1` was released to fix the problem. So probably should be `>= 2.1.1`.

gems/rwiki/CVE-2006-2581.yml
```diff
@@ -0,0 +1,39 @@
+---
+gem: rwiki
+cve: 2006-2581
+ghsa: gvhx-gj42-m28v
+url: https://web.archive.org/web/20090501134922/http://www2a.biglobe.ne.jp/~seki/ruby/rwiki.html
+title: RWiki before 2.1.0 has cross-site scripting vulnerability
```
postmodern, line 6: Also should be `2.1.1`.

postmodern (changes requested):
Google Translation of the page says the patched version is 2.1.1.

gems/rwiki/CVE-2006-2581.yml
jasnow: Fixed

gems/rwiki/CVE-2006-2581.yml
jasnow: fixed

jasnow:
thanks

PR #631, opened 2023-06-11
2nd GHSA rwiki advisories from 2006

jasnow:
2nd GHSA rwiki advisories from 2006
 * CVE-2006-2582.yml
 
 Previous one was CVE-2006-2581.


postmodern (approved):

jasnow:
thanks

PR #634, opened 2023-06-16
Added vapid-related URLs to older advisories

jasnow:
Added vapid and vapidlabs related URLs to older advisories.

Had to be extra careful for brbackup and kcapifony gems.
Verified that osvdb: number in new URLs was the same as existing osvdb: number.
Ran yamllint and rake (both green).


postmodern (approved):
LGTM

jasnow:
Thanks

PR #639, opened 2023-06-18
Added lint check to not support 'osvdb' in main url: field

jasnow:
Added lint check to not support 'osvdb' in main url: field since it is shuttered.

### DIFF
--- a/spec/advisory_example.rb
+++ b/spec/advisory_example.rb
@@ -94,6 +94,7 @@ shared_examples_for 'Advisory' do |path|
       subject { advisory['url'] }
 
       it { expect(subject).to be_kind_of(String) }
+      it { expect(subject).not_to include("osvdb") }
       it { expect(subject).not_to be_empty }
     end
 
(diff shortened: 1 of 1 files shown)


postmodern (changes requested):
This might erroneously flag URLs that just contain the string `osvdb`. You probably want to do `.not_to start_with("https://osvdb.org")`.

postmodern (commented):
I'm guessing you used `include()` instead of `start_with()` in order to catch `web.archive.org` URLs that load `http://osvdb.org`? `.to_not start_with()` would be more correct if we just want to filter out `http://osvdb.org/...` URLs.

jasnow:
Also I was not sure if it was http or https - recently I ran "git grep osvdb.org |grep http" and it is only "http" in repo.

postmodern:
`.to_not match(%r{\Ahttp(s)?://osvdb\.org})` would catch both `http://osvdb.org` and `https://osvdb.org`.

jasnow:
I liked my regexp.

jasnow:
Made your suggested change.

jasnow:
Thanks

PR #640, opened 2023-06-19
Removed 2 blank lines before debug info

jasnow:
Removed 2 blank lines before debug info so it is easier to remove debug info by script.


postmodern (commented):
What's the purpose of removing this or keeping it in? These empty-lines seem to just break up the boilerplate YAML from the GitHub Advisory YAML.

jasnow:
> What's the purpose of removing this or keeping it in? These empty-lines seem to just break up the boilerplate YAML from the GitHub Advisory YAML.

As I stated above "easier to remove debug info by script".

PR #643, opened 2023-06-21
Added 2016 gems/bundler/CVE-2016-7954.yml advisory

jasnow:
Added 2016 gems/bundler/CVE-2016-7954.yml advisory.

NOTE: This will flag all 1.x versions of bundler.

### DIFF
--- a/gems/bundler/CVE-2016-7954.yml
+++ b/gems/bundler/CVE-2016-7954.yml
@@ -11,12 +11,12 @@ description: |
   code into an application by leveraging a Gem name collision on a
   secondary source.
 
-  '*****************************************************************
-  Please note that this vulnerability only applies for Ruby projects
-  using Bundler < 2.0 with Gemfiles having 2 or more "source" lines.
-  *****************************************************************'
+  Please note that this vulnerability only applies for Ruby
+  projects using Bundler < 2.0 with Gemfiles having 2 or more
+  "source" lines.
 
-  NOTE: This might overlap CVE-2013-0334.
+  In other words, if the user's Gemfile does not use multiple
+  sources, this vulnerability can be ignored.
 cvss_v2: 7.5
 cvss_v3: 9.8
 patched_versions:
@@ -37,4 +37,4 @@ related:
     - https://github.com/rubygems/bundler/issues/5274
     - https://github.com/rubygems/bundler/issues/5051
     - https://github.com/rubygems/bundler/issues/5062
-notes: "GHSA is unreviewed"
+notes: "NOTE: This might overlap CVE-2013-0334.; GHSA is unreviewed"
--- a/gems/bundler/CVE-2016-7954.yml
+++ b/gems/bundler/CVE-2016-7954.yml
@@ -1,0 +1,35 @@
+---
+gem: bundler
+cve: 2016-7954
+ghsa: jvgm-pfqv-887x
+url: https://collectiveidea.com/blog/archives/2016/10/06/bundlers-multiple-source-security-vulnerability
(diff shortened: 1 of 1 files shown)


postmodern:
@indirect while I agree with you this is one of those vulnerabilities that requires a specific scenario to trigger, it is still a vulnerability and got a CVE. Also, Bundler 2.x has been out for a while, so maybe the annoyance of flagging Bundler 1.x as vulnerable might not be that bad? Also, why couldn't a patch be backported to Bundler 1.x? Unfortunately this is kind of what happens when a patch isn't backported, the whole previous version range has to be considered vulnerable.

@jasnow what do you think about adding an extra paragraph to the description, emphasizing that this vulnerability requires two `source`s be set in the `Gemfile`, otherwise users can safely ignore the vulnerability?

jasnow:
> @indirect while I agree with you this is one of those vulnerabilities that requires a specific scenario to trigger, it is still a vulnerability and got a CVE. Also, Bundler 2.x has been out for a while, so maybe the annoyance of flagging Bundler 1.x as vulnerable might not be that bad? Also, why couldn't a patch be backported to Bundler 1.x? Unfortunately this is kind of what happens when a patch isn't backported, the whole previous version range has to be considered vulnerable.
> 
> @jasnow what do you think about adding an extra paragraph to the description, emphasizing that this vulnerability requires two `source`s be set in the `Gemfile`, otherwise users can safely ignore the vulnerability?

Sure

jasnow:
What's a global source vs. "source 'https://rubygems.org' "?

jasnow:
"global source" defined here: https://bundler.io/v2.4/man/gemfile.5.html

gems/bundler/CVE-2016-7954.yml
```diff
@@ -11,6 +11,11 @@ description: |
   code into an application by leveraging a Gem name collision on a
   secondary source.
 
+  '*****************************************************************
+  Please note that this vulnerability only applies for Ruby projects
+  using Bundler < 2.0 with Gemfiles having 2 or more "source" lines.
+  *****************************************************************'
+
   NOTE: This might overlap CVE-2013-0334.
```
postmodern, line 19: I think that should be a `note: ` attribute and not part of the `description:` text.

gems/bundler/CVE-2016-7954.yml
```diff
@@ -11,6 +11,11 @@ description: |
   code into an application by leveraging a Gem name collision on a
   secondary source.
 
+  '*****************************************************************
+  Please note that this vulnerability only applies for Ruby projects
+  using Bundler < 2.0 with Gemfiles having 2 or more "source" lines.
+  *****************************************************************'
```
postmodern, line 17: No need for the asterisks, just a simple `Note: ` should suffice.

postmodern, line 17: Maybe also explicitly mention that if the user's `Gemfile` does not use multiple sources, the vulnerability can be ignored.

postmodern (changes requested):
Wording changes.

gems/bundler/CVE-2016-7954.yml
jasnow: Fixed

gems/bundler/CVE-2016-7954.yml
jasnow: Removed asterisks.

jasnow: Added text

postmodern (approved):

jasnow:
Thanks

PR #644, opened 2023-06-22
Add 'Prefer CVE filenames to GHSA filenames" lint rule and renamed 4 advisories

jasnow:
Add 'Prefer CVE filenames to GHSA filenames" lint rule and renamed 4 advisories.
For the one with 2 CVE's, I used the cve: value for filename and put both under related:/cve: field.


postmodern (approved):
Great idea.

PR #646, opened 2023-06-23
Updated info in sup gem advisories in support for issue #164

jasnow:
Updated info in sup advisories in support for issue #164 .


jasnow:
Thanks

PR #652, opened 2023-06-24
Fixed framework field method to work - which fixed 11 existing advisories

jasnow:
Fixed framework field method to work - which fixed 11 existing advisories.


postmodern (approved):
:+1: In the future we may want to refactor this method to use a constant of gem names to framework names. For now this is acceptable.

PR #654, opened 2023-06-25
Added cvss_v4 field

jasnow:
I know this is early, but wanted to at least have the field when we start getting the data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,10 +158,9 @@ patched_versions:
 * `related` \[Hash\<Array\<String\>\>\] (optional): Sometimes an advisory references many urls and other identifiers. Supported keys: `cve`, `ghsa`, `osvdb`, and `url`
 * `notes` \[String\] (optional): Internal notes regarding the vulnerability's inclusion in this database.
 
-[CVSSv2.0]: https://www.first.org/cvss/v2/guide
-[CVSSv3.0]: https://www.first.org/cvss/v3.0/user-guide
-[CVSSv3.1]: https://www.first.org/cvss/v3.1/user-guide
-[CVSSv4.0]: https://www.first.org/cvss/v4.0/user-guide
+[CVSSv2]: https://www.first.org/cvss/v2/guide
+[CVSSv3]: https://www.first.org/cvss/v3.1/user-guide
+[CVSSv4]: https://www.first.org/cvss/v4.0/user-guide
 
 ### Tests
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -157,7 +159,9 @@ patched_versions:
 * `notes` \[String\] (optional): Internal notes regarding the vulnerability's inclusion in this database.
 
 [CVSSv2]: https://www.first.org/cvss/v2/guide
-[CVSSv3]: https://www.first.org/cvss/user-guide
+[CVSSv3]: https://www.first.org/cvss/v3.0/user-guide
+[CVSSv3]: https://www.first.org/cvss/v3.1/user-guide
+[CVSSv3]: https://www.first.org/cvss/v4.0/user-guide
```
postmodern, line 164: You forgot to update the `v3` part of the link name.

spec/advisory_example.rb
```diff
@@ -154,6 +154,13 @@
         expect(subject).to be_kind_of(Float).or(be_nil)
       end
 
+      case advisory['cvss_v4']
+      when Float
+        context "when a Float" do
+          it { expect(subject).to be_between(0.0, 10.0) }
```
postmodern, line 160: Should be `advisory['cvss_v4']` or this logic needs to be moved into it's own `describe` block outside of the `describe "cvss_v3"` block.

README.md
```diff
@@ -157,7 +159,9 @@ patched_versions:
 * `notes` \[String\] (optional): Internal notes regarding the vulnerability's inclusion in this database.
 
 [CVSSv2]: https://www.first.org/cvss/v2/guide
-[CVSSv3]: https://www.first.org/cvss/user-guide
+[CVSSv3]: https://www.first.org/cvss/v3.0/user-guide
+[CVSSv3]: https://www.first.org/cvss/v3.1/user-guide
```
postmodern, line 163: We should probably only link to CVSS v3.1 when mentioning CVSSv3.

postmodern (changes requested):
Some mistakes I noticed.

README.md
jasnow: Add ".0" and ".1" to make it more specific.

README.md
jasnow: fixed

spec/advisory_example.rb
jasnow: The previous code also mentioned v2 in this block so I was confused.

jasnow: Made a separate describe block but did include non-v4 code from v3 code.

jasnow:
Thanks

README.md
postmodern: The text still uses the `[CVSSv3]` link name. That should be changed to `[CVSSv3.1]`. We don't link to `[CVSSv3.0]` anywhere so that link can be removed.

README.md
```diff
@@ -220,7 +220,9 @@ developed by the Open Security Foundation (OSF) and its contributors.
 [CVE]: https://cve.mitre.org/
 [OSVDB]: https://en.wikipedia.org/wiki/Open_Source_Vulnerability_Database
 [GHSA]: https://help.github.com/en/articles/about-maintainer-security-advisories
-[CVSSv2]: https://www.first.org/cvss/v2/guide
-[CVSSv3]: https://www.first.org/cvss/user-guide
+[CVSSv2.0]: https://www.first.org/cvss/v2/guide
+[CVSSv3.0]: https://www.first.org/cvss/v3.0/user-guide
+[CVSSv3.1]: https://www.first.org/cvss/v3.1/user-guide
```
postmodern, line 225: The README already uses `[CVSSv3]` link names, so either we keep the name or update the README to use `[CVSSv3.1]`.

postmodern (changes requested):
Need to make sure the README doesn't have broken/missing markdown links to the CVSS specs.

README.md
jasnow: fixed

jasnow:
Thanks

Issue #647, opened 2023-06-23
[Pre537] Add an explicit ignore list to the  github_advisory_sync.rb script


jasnow:
Add an explicit ignore list to the  github_advisory_sync.rb script.

Duplicates because of mixed case Gem names, Use CVE instead 
of GHSA filename, and Ruby => Gem reclassification.

Current list of files to be exclude: (6/23/2023)
 * ruby-advisory-db/gems/arabic-prawn/CVE-2014-2322.yml
 * ruby-advisory-db/gems/redcloth/CVE-2012-6684.yml
 * ruby-advisory-db/gems/spree_auth_devise/GHSA-6mqr-q86q-6gwr.yml
 * ruby-advisory-db/gems/spree_auth_devise/GHSA-8xfw-5q82-3652.yml
 * ruby-advisory-db/gems/spree_auth_devise/GHSA-gpqc-4pp7-5954.yml
 * ruby-advisory-db/gems/user_agent_parser/GHSA-pcqq-5962-hvcw.yml
 * ruby-advisory-db/gems/nokogiri/GHSA-fq42-c5rg-92c2.yml
 * ruby-advisory-db/gems/nokogiri/GHSA-gx8x-g87m-h5q6.yml
 * ruby-advisory-db/gems/nokogiri/GHSA-v6gp-9mmm-c6p5.yml
 * ruby-advisory-db/gems/nokogiri/GHSA-xxx9-3xcr-gjj3.yml
 * ruby-advisory-db/gems/webrick/CVE-2009-4492.yml

jasnow:
Closing this issue as this is covered by a script outside this repo.

PR #655, opened 2023-06-26
Shell-based solution to deleting duplicate advisories created by github_advisory_sync.rb script (#647)

jasnow:
Shell-based solution to deleting duplicate advisories created by github_advisory_sync.rb script. 
Part of issue #647.

Very open to safer or more ruby solutions, but this is simple and easy to maintain.


postmodern:
I'd prefer we just fix the code in `lib/github_advisory_sync.rb` to not create duplicates in the first place.

PR #658, opened 2023-06-29
Add "Prefer GHSA over OSVDB filename" lint rule

jasnow:
Added lint rule to check that OSVDB-* advisories with "ghsa:" values have filenames starting with GSHA.

To test this, add a "ghsa:" field/value to an existing OSVDB advisory and run "rake".


postmodern (approved):
Good idea.

PR #659, opened 2023-06-29
Corrected "ghsa:" values (they were duplicates) for 3 rubies/ruby advisories

jasnow:
Corrected "ghsa:" values (they were duplicates) for 3 rubies/ruby advisories.


postmodern (approved):
LGTM

Issue #649, opened 2023-06-24
[Pre537] Automate unaffected_versions conversion from GHSA to ruby-advisory-db


jasnow:
Automate unaffected_versions conversion from GHSA to ruby-advisory-db

jasnow:
### patched_version and unaffecte_verions Use Cases for Testing:
 * https://gist.github.com/jasnow/7c99f4d9e8b838293ba407af373141af

### (Raw) unaffected_versions Use Cases for Testing:
 * https://gist.github.com/jasnow/c075a3ecbabed38640b93a171b5700bf

jasnow:
Please review Roadmap above for missing unaffected_versions variables.

Is this true?
 * FYI: The only style rules regarding unaffected_versions field is be [double] quoted (CONTRIBUTING.md) and pass "rake" linting check.

postmodern:
Closed by PR #676.

PR #670, opened 2023-07-02
Singles: Working level step toward unaffected_versions field automation in support of #649

jasnow:
Singles: Working level step toward unaffected_versions: field automation in support of #649.
 * Mostly used code from patched_versions automation.
 * Now output: based on vulnerableVersionRange input. 
 * Now connected with the patched_versions values.
 * See TODO:'s for remaining known work.
 
"Singles" means a [single line, single range] for GHSA vulnerableVersionRange input.
See #949 for draft roadmap.

### DIFF
--- a/lib/github_advisory_sync.rb
+++ b/lib/github_advisory_sync.rb
@@ -382,7 +382,6 @@ module GitHub
     def unaffected_versions_for(package)
       # The unaffected_versions field is similarly not directly available.
       # This optional field must be inferred from the vulnerableVersionRange.
-
       # EXAMPLE YAML OUTPUT:
       #vulnerabilities:
       #- package:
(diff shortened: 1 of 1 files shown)


jasnow:
Ready for merge. Check out Roadmap in https://github.com/rubysec/ruby-advisory-db/issues/649 for more details.

postmodern (changes requested):
After looking over the algorithm, four things I'd like to do:

1. Generate the correct `unaffected_versions`, instead of generating pseudo-data which is then meant to be edited.
2. Simplify the logic. I think all we have to do is check if the first `vulnerableVersionRange` starts with `<=` vs. `>`/`>=`. If the first `vulnerableVersionRange` starts with `>` or `>=`, that means all versions below that version range are not vulnerable, thus unaffected. You could then derive the unaffected version range from the first `vulnerableVersionRange` by "inverting" the version range operator (ex: `> X.Y.Z` -> `<= X.Y.Z`, `>= X.Y.Z` -> `< X.Y.Z`).
3. Avoid using abbreviated method/variable names for readability. `fpv` or `vers` can be confusing to someone who's reading the code for the first time and hasn't yet learned the terminology.
4. Remove debugging `puts` statements and commented out code.

jasnow:
> After looking over the algorithm, four things I'd like to do:
> 
> 1. Generate the correct `unaffected_versions`, instead of generating pseudo-data which is then meant to be edited.
> 2. Simplify the logic. I think all we have to do is check if the first `vulnerableVersionRange` starts with `<=` vs. `>`/`>=`. If the first `vulnerableVersionRange` starts with `>` or `>=`, that means all versions below that version range are not vulnerable, thus unaffected. You could then derive the unaffected version range from the first `vulnerableVersionRange` by "inverting" the version range operator (ex: `> X.Y.Z` -> `<= X.Y.Z`, `>= X.Y.Z` -> `< X.Y.Z`).
> 3. Avoid using abbreviated method/variable names for readability. `fpv` or `vers` can be confusing to someone who's reading the code for the first time and hasn't yet learned the terminology.
> 4. Remove debugging `puts` statements and commented out code.

Your way of solving this problem will be better.

PR #687, opened 2023-07-18
Added no-pull-or-commit url: lint rule plus date-based phased introduction approach

jasnow:
Added no-pull-or-commit url: lint rule plus date-based phased introduction approach.

Expect to start fixing the 94 current advisories using the date-based 
phased introduction approach after this PR is merged.

The "date-based phased introduction approach" code 
will be removed after the 94 advisories are fixed.


postmodern:
I'd prefer to just fix all 94 advisories in a separate branch. I think this could be automated using either [`git sed`](https://github.com/postmodern/dotfiles/blob/a0560927e697c5d178504f86fb0da27fa0b47bf7/.gitconfig#L17) or a Ruby script to parse the YAML and then replace the `url:` line.

Also, the tests should check if the URL's host is `github.com` and the path contains `/pull/` or `/commit/`.

PR #695, opened 2023-08-15
GHSA SYNC: Added scripts/syncbot.sh to sponsor regular GHSA Sync task

jasnow:
GHSA SYNC: Added scripts/syncbot.sh to sponsor regular GHSA Sync task.

Green "rake" 

$ shellcheck myscript
No issues detected!


jasnow:
Current approach is probably better.

PR #696, opened 2023-08-18
Report puma GHSA-68xg-gqqm-vgj8 vulnerability

vlakre:
Add an entry for the new puma GHSA-68xg-gqqm-vgj8 vulnerability


postmodern:
Thank you!

PR #706, opened 2023-09-23
GHSA SYNC: Added 1 brand new advisory

jasnow:
GHSA SYNC: Added 1 brand new advisory: gems/mongrel/CVE-2007-6612.yml


postmodern (changes requested):
The primary `url:` should be `https://nvd.nist.gov/vuln/detail/CVE-2007-6612` since it explains the issue with mongrel. The Apple Security Bulletin lists multiple CVEs and you have to search within it to find where it mentions mongrel.

postmodern (approved):

PR #720, opened 2023-11-22
Add CVE-2019-8331 for `twitter-bootstrap-rails`

ddalcino:
CVE-2019-8331 has already been reported for the `bootstrap` and `bootstrap-sass` gems, but the `twitter-bootstrap-rails` gem is still affected, since it includes the vulnerable 3.3.6 version.

I'm not sure if it's appropriate for me to tell users that "hey, there's this other version of the bootstrap gem that's actually maintained by the original developer, and until this gem is fixed you should use it instead". I have done it anyway; please let me know if that needs to be changed.


postmodern (approved):

postmodern (changes requested):
Trailing whitespace on line 14 angered the yaml linter.

postmodern (approved):

PR #743, opened 2024-02-07
Update CVE-2021-41182, CVE-2021-41183, CVE-2021-41184

Saarett:
The Ruby gem `jquery-ui-rails` relies on the NPM package `jquery-ui` according [to this table](https://github.com/jquery-ui-rails/jquery-ui-rails/blob/master/VERSIONS.md).
It appears that the patched versions in `CVE-2021-41182`, `CVE-2021-41183`, and `CVE-2021-41184` are based on the NPM package `jquery-ui`. Therefore, this PR addresses the issue by adjusting them to reflect the versioning of `jquery-ui-rails`.

### DIFF
--- a/gems/jquery-ui-rails/CVE-2021-41182.yml
+++ b/gems/jquery-ui-rails/CVE-2021-41182.yml
@@ -35,7 +35,7 @@ description: |
 cvss_v2: 4.3
 cvss_v3: 6.5
 patched_versions:
-  - ">= 1.13.0"
+  - ">= 7.0.0"
 related:
   url:
     - https://github.com/jquery/jquery-ui/security/advisories/GHSA-9gj3-hwp5-pmwc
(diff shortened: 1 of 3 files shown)


postmodern (changes requested):
It doesn't look like `jquery-ui-rails` 7.0.0 has been released yet?
https://rubygems.org/gems/jquery-ui-rails/versions

Saarett:
@postmodern You're correct, I did miss it 🤔
Seems like for some reason the commit was tagged 7.0.0 but it wasn't released, and here is the thread about it: https://github.com/jquery-ui-rails/jquery-ui-rails/pull/139
The question is, do we rather set the patched version to nothing, or can this YAML file support a specific Github tag? Because you could require it this way:
`gem 'jquery-ui-rails', github: 'jquery-ui-rails/jquery-ui-rails', tag: 'v7.0.0'`

postmodern:
@Saarett double checked the `jquery-ui-rails` gem and they list the `jquery-ui.js` version in `lib/jquery/ui/rails/version.rb` (ex: `grep JQUERY_UI_VERSION lib/jquery/ui/rails/version.rb`).

I would still list `patched_versions` if a patched version is listed in the GHSA/CVE advisory. If no available gem version satisfies that patched version, then the developer can maybe bundle `jquery-ui-rails` from the GitHub repository.

postmodern (approved):

postmodern:
@Saarett double checked and confirmed the patched versions. I guess we could add a note to the `description:` that `7.0.0` is only currently available on GitHub.

PR #752, opened 2024-02-24
Moved to activestorage.

ooooooo-q:
The gem was incorrectly specified and has been corrected. (95316e8819722457f564263ee7fb448fbaee5063)


Rynaro (approved):
Based on the actual CVE-2024-26144, the `activestorage` is the fix target.

postmodern (approved):

PR #774, opened 2024-04-10

jasnow:
GHSA SYNC: 1 brand new advisory: gems/katello/CVE-2012-3503.yml

### DIFF
--- a/gems/katello/CVE-2012-3503.yml
+++ b/gems/katello/CVE-2012-3503.yml
@@ -1,0 +1,27 @@
+---
+gem: katello
+cve: 2012-3503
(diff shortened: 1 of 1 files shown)


postmodern:
The vulnerability was patched in commit https://github.com/Katello/katello/commit/1fd91b1dc855ba4b593f3d4c3fe3a850c4e79f23, which was tagged by the `katello-1.0.6-1` and `katello-1.1.7-1` release tags.

However, the first gem version of katello published to https://rubygems.org/ is 1.5.0. I suspect that prior to the katello 1.5.0 gem, katello was installed directly from git.

PR #784, opened 2024-06-02
add CVE-2024-34341 for ActionText

ddalcino:
Fix #783.

I hope this works for you; if not, I'm happy to apply changes.

Issue #783, opened 2024-05-31
CVE-2024-34341: XSS vulnerabilities in Trix editor


ddalcino:
Since Rails 7.0.8.2 & 7.0.8.3 were released on May 17th, a fix has been made available for this: CVE-2024-34341, https://discuss.rubyonrails.org/t/xss-vulnerabilities-in-trix-editor/85803

I've been waiting for this to show up in my bundle audit results; I figured this would get into the database via a standard GHSA sync. But I guess it hasn't yet, so here I am.

If the GHSA sync script doesn't generate new files for this, I'm happy to write one myself.

postmodern:
It doesn't appear that a CVE was never registered for [actiontext](https://rubygems.org/gems/actiontext) which embeds a copy of the Trix Editor, which is why we haven't picked it up yet.

flavorjones:
I believe the original intention was for the Trix GHSA to include Rails as "affected software", so I think adding an advisory under `gems/actiontext` is the right decision. I've pinged the Rails security team to let them know that Rails wasn't included in the GHSA.

PR #786, opened 2024-06-03
Correct affected versions for CVE-2024-34341

chadlwilson:
Amends #784 (earlier fix to #783) to match affected versions with that officially documented at https://discuss.rubyonrails.org/t/xss-vulnerabilities-in-trix-editor/85803

I can't find any source for the assertion that < 7.0 is vulnerable in contradiction to the Rails teams' assertions, so suggest removing it.


postmodern (approved):

chadlwilson:
Thanks, looks good! :-)

PR #846, opened 2025-01-10

jasnow:


jasnow:
Replaced with later PR.

PR #847, opened 2025-02-07
Remove Rails LTS versions

NiklasHae:
We (the Rails LTS team) noticed that some CVEs are marked as fixed by Rails LTS in the ruby-advisory-db. While we appreciate and are grateful for the effort, we would prefer not to list our commercial versions in the official database as fixed versions.

Currently, we do not have the capacity to actively monitor all versions listed in the ruby-advisory-db, verify their accuracy, and ensure correctness. Additionally, some fixes require extra context or additional information that may not be adequately reflected in the database. Most Rails LTS versions are a commercial offering, and we maintain a comprehensive list of all CVEs we have addressed—along with the necessary details—at https://makandracards.com/railslts/474590-list-cves-addressed-rails-lts. Every Rails LTS customer also receives a newsletter with updates and additional information for all patches.

We would prefer to keep our own documentation as the single source of truth for Rails LTS related fixes. Thank you for your understanding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ description: |
 
   Versions Affected: All
   Not affected: None
-  Fixed Versions: 5.2.8.15 (Rails LTS), 6.1.7.1, 7.0.4.1
+  Fixed Versions: 6.1.7.1, 7.0.4.1
 
   # Impact
 
@@ -85,7 +85,7 @@ description: |
 
   Users on Ruby 3.2.0 or greater are not affected by this vulnerability.
 patched_versions:
-  - "~> 5.2.8, >= 5.2.8.15"  # Rails LTS
+  - "~> 5.2.8"
   - "~> 6.1.7, >= 6.1.7.1"
   - ">= 7.0.4.1"
 ```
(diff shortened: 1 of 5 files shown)


postmodern:
First, I will need some kind of confirmation that this is indeed Niklas Häusele speaking on behalf of Rails LTS / Makandra, and not some random GitHub account possibly posing as Niklas Häusele.

Second, the Rails LTS patched versions were added by PR #538. If we remove the Rails LTS specific patched versions, then `bundle-audit` will flag those Rails LTS versions as vulnerable even though they've had patches backported to them. The user would then need to create or update their `.bunder-audit.yml` file and add CVEs to it as ignored. We should probably make a note of this in the README.

postmodern (changes requested):
Looks good, but also needs a note added to the README instructing Rails LTS users to check the official Rails LTS CVE list, and to add those CVEs to their `.bundler-audit.yml` file under `ignore:`.

jasnow:
Monitoring PR and will make changes to my GHSA SYNC scripts after the merge.

NiklasHae:
Thank you @postmodern for taking a look.

I sent you an email in your GitHub profile regarding verification. I'll update the PR accordingly :+1:

postmodern (approved):

postmodern:
Added a note to the README in eb74560.

PR #893, opened 2025-08-14
Some misc cleanup work

jasnow:
Some misc cleanup work
  * Upgraded activesupport gem to 8.0.x
  * Added "e.g" (example) text to README for framework: field.
  * Updated CVE-2014-10075 advisory to current standards. Removed dead links.


postmodern (changes requested):
These should be split these into separate commits.

PR #903, opened 2025-09-26
Added CKEditor 4 vulnerabilities

jamgregory:
Recent security scanning against one of our applications has highlighted a number of CKEditor 4 vulnerabilities we weren't originally aware of because they're not recorded in the advisory database.

I've brought these in by modifying the GitHub sync script slightly to pull in the security vulnerabilities listed on the https://github.com/ckeditor/ckeditor4 repository and then matching them against the gem version in https://github.com/galetahub/ckeditor.

Hopefully these are OK - please let me know if any further changes are required.

### DIFF
--- a/gems/ckeditor/CVE-2020-27193.yml
+++ b/gems/ckeditor/CVE-2020-27193.yml
@@ -1,0 +1,24 @@
+---
+gem: ckeditor
+cve: 2020-27193
+ghsa: 4m44-5j2g-xf64
+url: https://ckeditor.com/blog/CKEditor-4.15.1-with-a-security-patch-released/
+title: Improper Neutralization of Input During Web Page Generation in CKEditor4
+date: 2022-05-24
+description: |
+  A cross-site scripting (XSS) vulnerability in the Color Dialog plugin
+  for CKEditor 4.15.0 allows remote attackers to run arbitrary web script after persuading
+  a user to copy and paste crafted HTML code into one of editor inputs.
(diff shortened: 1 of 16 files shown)


postmodern (changes requested):
Ah ha. You are correct that versions < 5.0.0 contain a vendored version of ckeditor4. The `patched_versions:` will need to be updated accordingly, since technically `>= 5.0.0` is patched, as that was when the vendored `ckeditor4` was removed ([4.3.0](https://github.com/galetahub/ckeditor/tree/v4.3.0/vendor/assets/javascripts/ckeditor) vs. [5.0.0](https://github.com/galetahub/ckeditor/tree/v5.0.0/vendor/assets/javascripts/ckeditor)). (Although, I could be wrong, so feel free to correct me.)

jamgregory:
Just in case it's not clear (I suspect it's not :face_with_peeking_eye:), the Ruby library of `ckeditor` doesn't directly map to CKEditor versions (i.e. version 5.0 of the library points at version 4.11.3 of CKEditor 4 - https://github.com/galetahub/ckeditor/blob/v5.0.0/lib/ckeditor.rb#L75)

So... there's an assumption that could be made that the gem `>= 5.0.0` is patched, but if you don't customise the `cdn_url` option, you are still vulnerable up to `5.1.2` (`5.2.0` seems to switch to CKEditor 5 but hasn't been released yet)

postmodern (approved):
OK I see now.

PR #907, opened 2025-10-07
1 brand new advisory

jasnow:
1 brand new advisory: gems/uri/CVE-2025-61594.yml (based on ruby-lang post)


hudakh:
I'm new to this so sorry if this sounds obvious. Should this not be added to `rubies` as well? Some ruby versions would have the vulnerable version of the gem bundled.

https://www.ruby-lang.org/en/news/2025/10/23/ruby-3-3-10-released/

PR #934, opened 2025-12-16
Removed removed/duplicate URL

jasnow:
Removed removed/duplicate URL
 * https://github.com/rubysec/ruby-advisory-db/blob/master/gems/prosemirror_to_html/GHSA-52c5-vh7f-26fx.yml


postmodern (changes requested):
Appears that [GHSA-vfpf-xmwh-8m65](https://github.com/advisories/GHSA-vfpf-xmwh-8m65) has been withdrawn. We could remove the whole file.

postmodern:
I removed the file entirely since the [GHSA advisory](https://github.com/advisories/GHSA-vfpf-xmwh-8m65) was withdrawn.

Issue #674, opened 2023-07-04
[Pre537] Add success status response code check to lib/github_advisory_sync.rb for non-embedded URLs


jasnow:
[Pre537] Add success status response code check to lib/github_advisory_sync.rb for non-embedded URLs 

Currently best mitigation idea is to flag non-successful status response 
codes and manually replace them from https://archive.org/web data.

FIelds involved are:
 A. url:
 B. related:/url:
 ~~C. Embedded to description: field (strength goal)~~ (12/16/2026: Moved to #935)

jasnow:
Now (see #936) I see that this issue is not planned so closing it.

PR #936, opened 2025-12-16
Initial PR for issue #674

jasnow:
Here is the initial PR for this issue #674

Known assumptions
 * Only covers "url:" and "related:" (not "cve:" values) fields.
 * Assumes PR #934 has been merged.
 * Assumes http statuses (200, 301, 302, 308, and 403) are OKAY. Feedback is welcome.
 * Only covers advisories from 1/1/2025 going forward. Lots of cleanup with older advisories plus "rake" run is MUCH slower for all advisories.

### DIFF
--- a/spec/advisory_example.rb
+++ b/spec/advisory_example.rb
@@ -1,5 +1,6 @@
 require 'spec_helper'
 require 'versions_example'
+require 'net/http'
 
 require 'yaml'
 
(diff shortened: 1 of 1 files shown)


postmodern:
I previously decided to avoid checking the URLs as it would slow down the CI. A partial check might seem pragmatic, but then it allows dead URLs in older advisories.

jasnow:
See above comment @postmodern

PR #942, opened 2025-12-20
Fixed two sets of advisories with problem URLs

jasnow:
Fixed two sets of advisories with problem URLs. Details below.


jasnow:
Change in focus so deleting this PR

PR #945, opened 2026-01-01
GHSA SYNC: Added ghsa field/value to 1 advisory

jasnow:
GHSA SYNC: Added ghsa field/value to 1 advisory
  * gems/uri/CVE-2025-61594.yml

_FYI: Do them one at a time until I catch up._

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,11 +1,3 @@
-require 'strscan'
-
-class StringScanner
-  def peep(len)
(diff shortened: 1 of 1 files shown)


postmodern (changes requested):
I have pinned the CI Ruby version to 3.4 until Kwalify is fixed; or we decide to replace/remove it. You can remove the last two commits that fixes the CI and rebase against the master branch.

PR #949, opened 2026-01-08

jasnow:
 * new file:   gems/action_text-trix/GHSA-g9jg-w8vm-g96v.yml

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,11 +1,3 @@
-require 'strscan'
-
-class StringScanner
-  def peep(len)
-    peek(len)
-  end
-end
-
 require 'date'
 require 'kwalify'
 require 'rspec'


spec/spec_helper.rb
```diff
@@ -1,3 +1,11 @@
+require 'strscan'
+
+class StringScanner
+  def peep(len)
+    peek(len)
+  end
+end
```
postmodern: I have pinned the CI Ruby version to 3.4 until Kwalify is fixed; or we decide to replace/remove it. Please remove the `StringScanner` monkey patch.

postmodern (changes requested):

spec/spec_helper.rb
jasnow: Editted them  on GitHub

Issue #305, opened 2017-10-03
"encryptor" gem v2.0.0: AES-GCM nonce reuse vulnerability


tarcieri:
I'm not sure this vulnerability ever received a CVE. It's described in the project's README:

https://github.com/attr-encrypted/encryptor#upgrading-from-v200-to-v300

This gem was encrypting all messages using the same key/nonce. This not only exposes the XOR of the plaintexts if you XOR together two ciphertexts, but it also leaks the AES-GCM authentication key, allowing an attacker to forge messages and potentially perform chosen ciphertext attacks, which could potentially enable full plaintext recovery (especially if they've e.g. compromised the database, which is the threat model attr_encrypted is designed to defend against).

saghaulor:
More information on this vulnerability can be found here: https://github.com/attr-encrypted/encryptor/pull/22

jasnow:
Found this advisory:  https://advisories.gitlab.com/pkg/gem/encryptor/GMS-2016-16

jasnow:
NOTE: Found that this gem has not been released after fixing this vulnerability om 3.0.0.

jasnow:
Here is what I know about this issue (in advisory/yaml format) with notes:
* https://gist.github.com/jasnow/b035470d0d84c090d2adb5059f113df5

Since this repo does not create contents or CVEs/GHSAs, I will be **closing**
this issue shortly. Speak up if you want to work on this.

jasnow:
Acting on my previous comment, I am closing this issue.

PR #960, opened 2026-01-14
Add gsm: advisory field in support of issue #305; Added 1 brand new GSM advisory

jasnow:
Add gsm: advisory field in support of issue #305
  * Modified the schemas, specs, and README to add "gsm:" advisory field (similar to "ghsa:" field).
   * Add GSM-2016-16 advisory

### DIFF
--- a/spec/advisory_example.rb
+++ b/spec/advisory_example.rb
@@ -102,6 +102,7 @@ shared_examples_for 'Advisory' do |path|
       it "may be nil or a String" do
         expect(subject).to be_kind_of(String).or(be_nil)
       end
+
       it "should be id in filename if filename is GHSA-XXX" do
         if filename_ghsa
           expect(subject).to eq(filename_ghsa.chomp('.yml'))


postmodern:
I'm hesitant about adding GSM advisory IDs to the database because of one advisory that never got assigned a CVE. [bundler-audit](https://github.com/rubysec/bundler-audit) would also need to be updated, as it expects either a `cve:`, `ghsa:`, or `osvdb:` ID to be present; otherwise `encryptor` < 3.0.0 would break `bundler-audit`.

I reviewed [encryptor's GitHub issues](https://github.com/attr-encrypted/encryptor/issues)'s and noted two things:

1. Appears that the maintainer said they would [try to request a CVE](https://github.com/attr-encrypted/encryptor/issues/30#issuecomment-336960001), but never reported back.
2. Another person did [further research and believes the vulnerability is actually in Ruby's `openssl` library](https://github.com/attr-encrypted/encryptor/pull/22#issuecomment-202083079), not the `encryptor` gem. This might explain why a CVE was never assigned.

Either someone else needs to request the CVE on behalf of [encryptor](https://github.com/attr-encrypted/encryptor), or maybe a CVE is not needed and [GSM-2016-16](https://gitlab.com/gitlab-org/security-products/gemnasium-db/-/blob/master/gem/encryptor/GMS-2016-16.yml?ref_type=heads) might be invalid.

Long-term, I might be open to trying to consume GitLab's Advisory database along with GitHub's Advisory DB and NVD.

PR #992, opened 2026-02-12
Add patched version for CVE-2026-25765

toddsiegel:
The fix for this was backported to 1.x versions as well.

See: 
- lostisland/faraday@d0fc049beb
- https://github.com/advisories/GHSA-33mh-2634-fwr2

### DIFF
--- a/gems/faraday/CVE-2026-25765.yml
+++ b/gems/faraday/CVE-2026-25765.yml
@@ -61,7 +61,7 @@ description: |
   ```
 cvss_v3: 5.8
 patched_versions:
-  - ">= 1.10.5"
+  - "~> 1.10.5"
   - ">= 2.14.1"
 related:
   url:


gems/faraday/CVE-2026-25765.yml
```diff
@@ -61,14 +61,17 @@ description: |
   ```
 cvss_v3: 5.8
 patched_versions:
+  - ">= 1.10.5"
```
postmodern: Should be `~> 1.10.5`, otherwise it will incorrectly state that `>= 1.10.5, < 2.14.1` is patched.

postmodern (changes requested):
Only the last patched version number should start with a `>=`. The previous patched versions should use `~>`.

gems/faraday/CVE-2026-25765.yml
postmodern: Fixed it myself.

postmodern (approved):

PR #1001, opened 2026-02-18
GHSA sync script fixes

jamgregory:
Just made two amendments to the GHSA sync script to fix issues I noticed when using it recently:

1. Correct the generation of the vulnerabilities references

    Currently the references nest an additional `url:` key inside them, which causes the RSpec tests to fail

2. Translate two package names that appear to have been renamed at some point

    `arabic-prawn` and `redcloth` apparently previously used camel-cased names (`Arabic-Prawn` and `RedCloth`) which is how they're stored in the repository. The sync script creates 'new' vulnerabilities because of these changes, so I've translated these package names back to avoid this.

### DIFF
--- a/lib/github_advisory_sync.rb
+++ b/lib/github_advisory_sync.rb
@@ -414,7 +414,7 @@ module GitHub
 
       # populate the related information
       new_data["related"] = {
-        "url" => advisory["references"]
+        "url" => advisory["references"].map { |reference| reference['url'] }.reject(&:empty?)
       }
 
       FileUtils.mkdir_p(File.dirname(filename_to_write))
--- a/lib/github_advisory_sync.rb
+++ b/lib/github_advisory_sync.rb
@@ -184,6 +184,11 @@ module GitHub
 
   class GitHubAdvisory
     class Package
+      NORMALISED_NAMES = {
+        "arabic-prawn"  => "Arabic-Prawn",
+        "redcloth"      => "RedCloth",
+      }
+
       attr_reader :name
 
       def initialize(advisory, name)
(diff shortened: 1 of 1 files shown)


lib/github_advisory_sync.rb
```diff
@@ -196,7 +196,18 @@ def updating?
...
+        # These packages appear to have been named differently in the past
+        # This 'corrects' them so updates don't affect existing vulnerabilities
+        package_name = case name
+        when "arabic-prawn"
+          "Arabic-Prawn"
+        when "redcloth"
+          "RedCloth"
+        else
+          name
+        end
```
postmodern: This should be a `Hash` lookup.

lib/github_advisory_sync.rb
```diff
@@ -414,7 +425,7 @@ def create(package)
 
       # populate the related information
       new_data["related"] = {
-        "url" => advisory["references"]
+        "url" => advisory["references"].map { |reference| reference['url'] }.reject(&:empty?)
```
postmodern: Are you sure `references` contains `url`? `references` comes from GitHub's Advisory API.

jasnow:
Approve changes - Ready to merge.

lib/github_advisory_sync.rb
flavorjones: Instead of an inline case statement, I would prefer to introduce a constant hash for this. Something like:

```ruby
class Package
  NORMALIZED_NAMES = { "arabic-prawn" => "Arabic-Prawn" }

  def filename
    NORMALIZED_NAMES[name] || name
  end
end
```

which extracts the name changes into a well-named data structure that is easily extensible.

flavorjones (changes requested):

lib/github_advisory_sync.rb
simi: We can even extract this outside of the code into some simple file with mapping, so it can be updated on its own.

jamgregory: I've gone with the constant as the easiest change - happy to switch to using a file if that's preferred (a YAML file, perhaps?)

simi: I'm ok with, good to move file if we will see a lot of traffic there.

flavorjones:
Thank you!

Issue #1045, opened 2026-05-12
WIP: Local ruby-advisory-db conventions


jasnow:
Local advisory conventions that @postmodern and I (@jasnow) have used in the past and are not checked by "yamllint" or "rake" run. 

5/12/2026
   * The first step before submitting PRs is to run "yamllint" on the YAML files then run "rake". Both commands should find no issues. More info regarding yamllint [HERE](https://github.com/rubysec/ruby-advisory-db/blob/master/.github/workflows/ruby.yml).
   * See PR #1022 regarding checking for correct YAML field indentation.
   * Field `related:/url:` is 4 blanks from left margin.
   * Field `patched_versions` and `unaffected_versions` are 2 blanks from left margin.
  * Postmodern likes:
    * Change "|-" to "|" on description: line.
    * Line wrap `descriptions:` and `title:` field at 80. (See [CONTRIBUTING.md](https://github.com/rubysec/ruby-advisory-db/blob/master/.github/workflows/ruby.yml)). @jasnow usually uses 75 because of the 2-char field indent.

5/13/2025
  * Postmodern likes:
     * Not use "\n" in `description:` field.
     * No "POC" in `description:` field. They will be flagged during `rake` run and removed during harvesting.
   * ruby YAML does not like embedded ":" characters.
   * Check all URLs for dead links. Sometimes find the URL  https://web.archive.org .
   * Suggest adding project-related evidence as references to prove the patch, such as
     * CHANGELOGs, Release Notes, project blog posts.

5/15/2026
  * Postmodern wanted the advisory filename prefix to be named: 1st choice: CVE, then GHSA, then OSVDB.

5/23/2026
 * Regarding the PR reviews yesterday:
    * I (@jasnow) usually collects all of the necessary URLs and put them in the `related: / url` field 
       then pick one the above URL that is an advisory to use in the `url:` field. Never thought of it as
      duplicates and @postmodern never asked for the duplicate to be removed.
   * I found `dependabot` being used in 2023 but it stopped at some point. Also `.gitignore` file 
     contains Gemfile.lock file so @postmodern did not work gem upgrade PRs. 

5/24/2026:
  * @postmodern wants this policy: `Omit patched_versions: if the GHSA has no patched version identifiers` as documented in https://github.com/rubysec/ruby-advisory-db/pull/664/changes
      More info at: https://github.com/rubysec/ruby-advisory-db/issues/157
[...]

PR #1054, opened 2026-05-17
Json schema yaml validation

dennispaagman:
I mentioned using [JSON Schema](https://json-schema.org/) to validate / lint the yml files in the Bundler slack as a possible improvement to the workflows for this repo.

I've made a PoC  below to see what's possible. It already flagged a few inconsistenties that I also fixed, so that show's there some possible value here.

---

Right now there are a couple of rules defined in various places:

- Rules defined in the [README](https://github.com/rubysec/ruby-advisory-db/tree/master#yaml-schema).
- A [shared example](https://github.com/rubysec/ruby-advisory-db/blob/master/spec/advisory_example.rb) that is ran over all files that does some data checks using those rules.
- [kwalify](https://github.com/kvs/kwalify) linting. Note that this gem has been deprecated/archived and does not work with Ruby 4.x. 
- [yamllint](https://github.com/rubysec/ruby-advisory-db/blob/master/.yamllint.yml).
- [An issue with some best practices](https://github.com/rubysec/ruby-advisory-db/issues/1045).

This PR condenses those into two JSON schemas and adds a test to validate all files against the schemas by converting the yml into JSON-like ruby hashes. (_Disclaimer: Claude Code was used, but I checked all output. And I hand wrote this PR description 😂_)

JSON Schema is nice as it's more powerful in what you can define than kwalify does (like not allowing certain values).  it's supported in most modern editors (you can get autocomplete on values, warnings, etc working when editing files). In a way it can also work as a single source of truth for the rules.

---

There are a couple of considerations that I can think of from the top of my head:

- It compares values, not syntax of yml files, so yamllint is probably still needed for that.
- Editor support does not come out of the box. You need a magic comment to each yml file or set up it yourself.
- The transformation to Ruby objects is a bit clunky. There are JS based linters that can just run as is, but that would require introducing a JS toolchain into the repo.
- Maybe it's too strict? I don't have the experience to judge that for this repo.
- It needs to be added as a github action workflow step.
- Can we remove `kwalify`? There's quite a bit of duplication with that and the `advisory_*` specs. I left them for now.
- Maybe deduplicate the shared stuff in the schemas.

I've made two separate commits, one that adds the schemas and it's test and one that fixes the issues it flagged in existing files. I've fixed them as I think makes sense, but they can definitely use a second pair of eyes.

Curious to hear what you think, I just wanted to at least put some effort in to show what it could look like.


jasnow:
Review feedback **UPDATED**
 1. I am a main user of these checks so **more checks are better.** Personally prefer YAML over JSON.
  2. (now done in PR#1060) ~~Please add URL in RELATED/CVE and RELATED/CHSA into RELATED/URL field so we do not lose the data context~~
 3. Please keep the old "rake" checks (at least for a while).
Thanks for your contribution.

**Will move my suggested enhancements into separate issue.**

jasnow:
LGTM - Approved - Ready to merge.

gems/httparty/CVE-2025-68696.yml
```diff
@@ -55,44 +55,6 @@ description: |
   Also, Python's `urljoin` function has documented a warning about similar behavior:
   <https://docs.python.org/3.13/library/urllib.parse.html#urllib.parse.urljoin>
 
-  ## PoC
```
simi, line 58: Why is this part removed? 🤔

simi: Yeah, why this part of description was removed?

jasnow: @postmodern did want it - https://github.com/rubysec/ruby-advisory-db/pull/790

simi: OK, I would prefer to just keep original descriptions untouched. But all good.

simi (approved):

gems/httparty/CVE-2025-68696.yml
jasnow: > OK, I would prefer to just keep original descriptions untouched. But all good.

I agree.

PR #1063, opened 2026-05-23
New advisory - see Ruby 4.0.5

jasnow:
New advisory - see Ruby 4.0.5
 * rubies/ruby/CVE-2026-46727.yml

### DIFF
--- a/rubies/ruby/CVE-2026-46727.yml
+++ b/rubies/ruby/CVE-2026-46727.yml
@@ -21,7 +21,7 @@ description: |
   This issue has been fixed in Ruby 4.0.5. We recommend upgrading Ruby.
 cvss_v3: 8.1
 unaffected_versions:
-  - "<= 3.4"
+  - "< 4.0"
 patched_versions:
   - ">= 4.0.5"
 related:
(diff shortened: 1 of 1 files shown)


rubies/ruby/CVE-2026-46727.yml
```diff
@@ -0,0 +1,35 @@
...
+  Memory-corruption-based exploitation is theoretically possible. The
+  attack could, for example, be carried out through a crafted
+  authoritative DNS server or recursive resolver.
+
+  This vulnerability has been assigned the CVE identifier CVE-2026-46727.
+
+  This issue has been fixed in Ruby 4.0.5. We recommend upgrading Ruby.
+cvss_v3: 8.1
+unaffected_versions:
+  - "<= 3.4"
```
flavorjones: I think I would prefer this to be `< 4.0` for clarity. "3.4" without a patch version can be confusing when an "=" is used. (For example, it's not clear that "3.4.1" <= "3.4".)

jasnow: Added commit

flavorjones (approved):

PR #1067, opened 2026-05-23
Updated CI to Ruby 4.x

jasnow:
Updated CI to Ruby 4.x.
 * Add monkey patch to deal with the missing ``peep` function.


simi:
I think we need to remove kwalify (as in other PR) or resurrect it. It is not good idea to keep patches around in app inside.

jasnow:
Easy to say - hard to do.

simi (approved):

simi:
@jasnow let's merge this. I'll check on the migration out of deprecated deps on side.

PR #1068, opened 2026-05-24
Add text to README.md regarding unreleased patches

### LINKED ISSUE #157: Dealing with unfixed vulnerabilities in gems
What's the best way to handle OSVDB entries / CVE assignments for ruby gems with **unfixed** vulnerabilities? Specifically, maybe a gem is obsolete / unmaintained and won't ever have a new fixed version, but we want to let people know they are using a vulnerable gem. Another case is when a gem takes too long to fix an issue, but we want to warn users so they are aware (maybe not cause a failure, but at least a warning in those cases).

jasnow:
Add text to README.md regarding unreleased patches
 * Example: `Please omit the "patch_versions:" field and value if the patch has not been released.`
 
 Hopefully this will address/fix https://github.com/rubysec/ruby-advisory-db/issues/157 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,8 +135,10 @@ patched_versions:
 * `cvss_v4` \[Float\] (optional): The [CVSSv4] score for the vulnerability.
 * `unaffected_versions` \[Array\<String\>\] (optional): The version requirements for the
   unaffected versions of the Ruby library.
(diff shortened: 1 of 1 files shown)


jasnow:
These changes overlap with PR#1070 so closing as DUPLICATE.

PR #1069, opened 2026-05-24
New rake check - compare root url: to filename (without suffix)

jasnow:
Based on @simi's feedback during the past week or two, I tried to create a new "rake" lint check (rule) for root `url:` == basename(filenname).

 * First, since this is a new rule and the large number of failed checks on existing advisories, I decided to pick a start date to start applying it. I found that advisories with `date:` > "2026-05-09" all passed. I know this is a compromise but I hope the team agree that the new check is worth the compromise.

 * Second, I skipped the OSVDB advisories because they are very old and that database was removed from the web and internet archive.

### DIFF
--- a/spec/advisory_example.rb
+++ b/spec/advisory_example.rb
@@ -110,9 +110,6 @@ shared_examples_for 'Advisory' do |path|
       it "has a filename that matches the root of the url field" do
         url = advisory["url"]
 
-        # Extract last path segment from URL
-        url_root = File.basename(URI.parse(url).path)
-
         # Extract filename without extension
         filename_root = File.basename(path, ".yml")
 
@@ -120,7 +117,7 @@ shared_examples_for 'Advisory' do |path|
         start_date = Date.new(2026, 5, 9)
         # Skip advisories older than start_date and old OSVDB advisories.
         if advisory["date"] >= start_date and !filename_root.start_with?("OSVDB")
-          expect(filename_root).to eq(url_root),
+          expect(url).to include(filename_root),
             "Expected filename DOES NOT to match URL root"
         end
       end
--- a/spec/advisory_example.rb
+++ b/spec/advisory_example.rb
@@ -110,15 +110,12 @@ shared_examples_for 'Advisory' do |path|
       it "has a filename that matches the root of the url field" do
         url = advisory["url"]
 
-        # Extract filename without extension
         filename_root = File.basename(path, ".yml")
 
         # 5/24/2026: May 9, 2026 is earliest start date with no failed checks.
         start_date = Date.new(2026, 5, 9)
-        # Skip advisories older than start_date and old OSVDB advisories.
         if advisory["date"] >= start_date and !filename_root.start_with?("OSVDB")
-          expect(url).to include(filename_root),
-            "Expected base filename DOES NOT include URL"
+          expect(url).to include(filename_root)
         end
       end
     end
--- a/spec/advisory_example.rb
+++ b/spec/advisory_example.rb
@@ -121,7 +121,7 @@ shared_examples_for 'Advisory' do |path|
         # Skip advisories older than start_date and old OSVDB advisories.
         if advisory["date"] >= start_date and !filename_root.start_with?("OSVDB")
           expect(filename_root).to eq(url_root),
-            "Expected filename '#{filename_root}' DOES NOT to match URL root '#{url_root}'"
+            "Expected filename DOES NOT to match URL root"
         end
       end
     end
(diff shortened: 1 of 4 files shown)


jasnow:
Hint: If you want to see all the failed check, change the **start date to 2000** and you will see all of them.

spec/advisory_example.rb
```diff
@@ -106,6 +106,24 @@
       it { expect(subject).to be_kind_of(String) }
       it { expect(subject).to_not match(%r{\Ahttp(s)?://osvdb\.org}) }
       it { expect(subject).not_to be_empty }
+
+      it "has a filename that matches the root of the url field" do
+        url = advisory["url"]
+
+        # Extract last path segment from URL
+        url_root = File.basename(URI.parse(url).path)
```
flavorjones: I wonder if it would be useful to instead assert something like

```
expect(url).to include(filename_root)
```

resulting in a message like

```
expected "https://github.com/lsegal/yard/security/advisories/GHSA-3jfp-46x4-xgfj" to include "CVE-2026-41493"
```

Then the assertion would pass on URLs like `https://www.cve.org/CVERecord?id=CVE-2026-45363` (see "/gems/jwt/CVE-2026-45363.yml").

spec/advisory_example.rb
```diff
@@ -106,6 +106,24 @@
...
+
+        # Extract filename without extension
+        filename_root = File.basename(path, ".yml")
+
+        # 5/24/2026: May 9, 2026 is earliest start date with no failed checks.
+        start_date = Date.new(2026, 5, 9)
+        # Skip advisories older than start_date and old OSVDB advisories.
+        if advisory["date"] >= start_date and !filename_root.start_with?("OSVDB")
+          expect(filename_root).to eq(url_root),
+            "Expected filename '#{filename_root}' DOES NOT to match URL root '#{url_root}'"
```
flavorjones: This string is not used as the error message. RSpec will generate the right message based on the assertion. It can be removed.

spec/advisory_example.rb
```diff
@@ -106,6 +106,24 @@
...
+        url = advisory["url"]
+
+        # Extract last path segment from URL
+        url_root = File.basename(URI.parse(url).path)
+
+        # Extract filename without extension
+        filename_root = File.basename(path, ".yml")
+
+        # 5/24/2026: May 9, 2026 is earliest start date with no failed checks.
+        start_date = Date.new(2026, 5, 9)
```
flavorjones, line 116: I like this approach. I've heard this approach called "ratcheting", where you set a linting policy that is immediately in effect for new code/reports, but allows older code to pass; and then have the ability to then gradually apply it to the rest of the codebase/reports.

spec/advisory_example.rb
jasnow: Added commit.

spec/advisory_example.rb
jasnow: add commit

jasnow:
working on failed GHA CI for 2 advisory from yesterday

jasnow:
@flavorjones - Now it is ready for re-review.

spec/advisory_example.rb
```diff
@@ -106,6 +106,21 @@
...
+        # Skip advisories older than start_date and old OSVDB advisories.
+        if advisory["date"] >= start_date and !filename_root.start_with?("OSVDB")
+          expect(url).to include(filename_root),
+            "Expected base filename DOES NOT include URL"
```
flavorjones: The "Expected base filename ..." string doesn't do anything here, and should be dropped.

spec/advisory_example.rb
```diff
@@ -106,6 +106,21 @@
       it { expect(subject).to be_kind_of(String) }
       it { expect(subject).to_not match(%r{\Ahttp(s)?://osvdb\.org}) }
       it { expect(subject).not_to be_empty }
+
+      it "has a filename that matches the root of the url field" do
+        url = advisory["url"]
+
+        # Extract filename without extension
```
flavorjones: This comment is probably not necessary, the code is pretty clear.

spec/advisory_example.rb
```diff
@@ -106,6 +106,21 @@
...
+        filename_root = File.basename(path, ".yml")
+
+        # 5/24/2026: May 9, 2026 is earliest start date with no failed checks.
+        start_date = Date.new(2026, 5, 9)
+        # Skip advisories older than start_date and old OSVDB advisories.
```
flavorjones: This comment is also probably unnecessary, the code reads well.

spec/advisory_example.rb
jasnow: done

spec/advisory_example.rb
jasnow: done

spec/advisory_example.rb
jasnow: done

PR #1075, opened 2026-05-26
Change Rakefile to run json checks first then all checks

jasnow:
Change Rakefile to run json checks first then all checks
  * Rakefile
  
This will help isolate issue with the checks going forward.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,7 +15,7 @@ end
 desc "Sync GitHub RubyGem Advisories into this project"
 task :sync_github_advisories, [:gem_name] do |_, args|
   require_relative "lib/github_advisory_sync"
-  t.exclude_pattern = "spec/schema_validation_spec.rb'
+  t.exclude_pattern = "spec/schema_validation_spec.rb"
   GitHub::GitHubAdvisorySync.sync(gem_name: args[:gem_name])
 end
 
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -18,5 +18,9 @@ task :sync_github_advisories, [:gem_name] do |_, args|
   GitHub::GitHubAdvisorySync.sync(gem_name: args[:gem_name])
 end
 
+RSpec::Core::RakeTask.new(:schema) do |t|
+  t.pattern = 'spec/schema_validation_spec.rb'
```
flavorjones: This file is already included by the default pattern used for the `:lint` check, meaning that if I run `bundle exec rake` it will run the schema tests twice.

You may want to declare the `:yaml` task with an `exclude_pattern` so that this spec file is only handled by one or the other (but not by both).

jasnow: Add the `exclude_pattern` but did not see the `:yaml" task comment until after I finished. Hopefully we can do that separately if needed.

jasnow:
@flavorjones - ready for re-review

flavorjones:
@jasnow I pushed a commit that cleans up the Rakefile (which was structured in a way that was making this change more difficult than it needed to be). Let me know what you think, if you like it I'll merge.

jasnow:
LGTM

jasnow:
Thanks

PR #1079, opened 2026-05-28
Add a rspec lint check to check title: and description:line lengths.

jasnow:
Add a rspec lint check to check `title:` and `description:`line lengths.
 * Initial the maximum `title:`             line length will be 154.
 * Initial the maximum `description:` line length will be 80.
 * Initial start date is May 9, 2026 and it is the earliest start date with no failed checks.
  * Also excluded OSVDB advisory since they are very old.
  
These max numbers are at or close to the minimum to get a clean "rake" run.

### DIFF
--- a/spec/advisory_example.rb
+++ b/spec/advisory_example.rb
@@ -3,6 +3,9 @@ require 'versions_example'
 
 require 'yaml'
 
+MAX_DESC_LEN  = 80
+MAX_TITLE_LEN = 154
+
 shared_examples_for 'Advisory' do |path|
   advisory = YAML.safe_load_file(path, permitted_classes: [Date])
 
@@ -106,6 +109,18 @@ shared_examples_for 'Advisory' do |path|
       it { expect(subject).to be_kind_of(String) }
       it { expect(subject).to_not match(%r{\Ahttp(s)?://osvdb\.org}) }
       it { expect(subject).not_to be_empty }
+
+      it "has a filename that matches the root of the url field" do
+        url = advisory["url"]
+
+        filename_root = File.basename(path, ".yml")
+
+        # 5/24/2026: May 9, 2026 is earliest start date with no failed checks.
+        start_date = Date.new(2026, 5, 9)
+        if advisory["date"] >= start_date and !filename_root.start_with?("OSVDB")
+          expect(url).to include(filename_root)
+        end
+      end
     end
 
     describe "title" do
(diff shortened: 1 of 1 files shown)


flavorjones:
@jasnow Did you explore whether it's easier to add these checks to `spec/schemas/gem.json` now that we have the json schemea validator? I suspect this is something it can do quite easily (and quickly).

simi:
@jasnow good to keep in json scheme as much as possible IMHO. It is easier to maintain and easier for people to contribute.

StantonMatt (commented):
I ran this locally with Ruby 3.3.11:

```console
bundle exec rake lint
61543 examples, 0 failures

bundle exec rspec spec/schema_validation_spec.rb
1279 examples, 0 failures
```

`git diff --check` also passes.

On the schema question: I think the title/description checks can mostly live in JSON Schema if the project wants them there. `title` can use `maxLength`, and the description line limit can be expressed with a `not` pattern against lines of 81+ characters. Since these shared examples apply to all advisories, that would probably need to be added to both `spec/schemas/gem.json` and `spec/schemas/ruby.json`, not only the gem schema.

The caveat is the grandfathering. I tried adding `title.maxLength = 154` and a description long-line `not` pattern to the gem schema locally, and schema validation reports 467 existing gem advisory failures. So the date/OSVDB gate in this PR is doing real work unless maintainers want to clean up existing records or enforce the rule globally.

The filename-root URL check is a different case: the current JSON Schema validation only sees the advisory data, not the file path, so that rule does not map cleanly to the existing schema validator. But that check already exists on current `master`, so I would keep this PR focused on the two length rules.

Given that, I think either direction is reasonable:

- keep this as RSpec logic if the goal is incremental enforcement for newer advisories only;
- move the two length limits into JSON Schema if the project wants the simpler long-term rule and is willing to handle the existing-data failures.

Small cleanup if this stays in RSpec: the description example string currently uses `${MAX_DESC_LEN}` instead of Ruby interpolation, and that block has a few indentation inconsistencies.

jasnow:
> @jasnow good to keep in json scheme as much as possible IMHO. It is easier to 
> maintain and easier for people to contribute.

 Other people was welcome to contribute in JSON.

jasnow:
> @jasnow Did you explore whether it's easier to add these checks to `spec/schemas/gem.json` now that we have the json schemea validator?

I have looked a JSON in the past but found it not as compact as YAML so I wrote my checks in YAML.
Looking forward to using this during my regular GHSA/SYNC work.

jasnow:
IInteresting: https://github.com/ducks/yaml-janitor

jasnow:
I guess it is time to close this proposed enhancement - thought it was a good compromise.

jasnow:
Json check did not catch long lines so reopening this PR.

jasnow:
Think it will be easier to create a new PR so closing this one.

PR #1080, opened 2026-05-30
Added .ruby-version to .gitignore file

jasnow:
Added .ruby-version to .gitignore file because Gemfile.lock was already in it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /Gemfile.lock
+.ruby-version
 .bundle
 .DS_Store
 /vendor


simi:
@jasnow Just wondering - If not needed, why to bother with in .gitignore? Do you use it locally and is it just causing issues in git status?

jasnow:
Added it to my global .gitignore file.

simi:
> Added it to my global .gitignore file.

I'm happy to add here if that helps you. Just wondering. Adding to .gitignore could cause issues to other projects you would like to keep the file. :thinking:

jasnow:
> I'm happy to add here if that helps you. Just wondering. Adding to .gitignore could
> cause issues to other projects you would like to keep the file. 🤔

I'm okay with new approach for now. Will create another ticket if it become a problem.

PR #1084, opened 2026-05-30
Add nekohtml advisory based on nokogiri advisory

jasnow:
Add nekohtml advisory based on nokogiri advisory
  *  gems/nekohtml/CVE-2022-24839.yml

### DIFF
--- a/gems/nekohtml/CVE-2022-24839.yml
+++ b/gems/nekohtml/CVE-2022-24839.yml
@@ -1,0 +1,35 @@
+---
+gem: nekohtml
+platform: jruby
+cve: 2022-24839
+ghsa: gx8x-g87m-h5q6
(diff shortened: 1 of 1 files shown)


flavorjones:
I personally don't think it's necessary to create this file.

1. The Nokogiri fork of `nekohtml` is not available as a ruby gem.
2. It is ONLY vendored in Nokogiri, and `gems/nokogiri/CVE-2022-24839.yml` already exists to notify people to upgrade Nokogiri.

So I don't know who's going to use this information, or how it will be actionable.

I recommend closing without merging, but will defer to the other maintainers if they feel strongly it should be part of the record.

jasnow:
Based on @flavorjones and @simi feedback, I will closing this PR.

PR #1085, opened 2026-05-31
Move docs to better markdown structure.

simi:
- fix typo in filename
- text should be 1:1, just improved formatting a little


jasnow (approved):
LGTM

PR #1087, opened 2026-05-31
Removing OSVDB-118481.yml

jasnow:
Based on the answer in https://github.com/rubysec/ruby-advisory-db/discussions/1082l,
I am deleting OSVDB-118481.yml from the database.


simi (approved):

jasnow:
thanks

PR #1092, opened 2026-06-02
Fix GitHub advisory YAML indentation

### LINKED ISSUE #1091: Fix `lib/github_advisory_sync.rb` script to indent `patched_versions:` and `related:/url:` 2 spaces to the right
Fix `lib/github_advisory_sync.rb` script to indent `patched_versions:` and `related:/url:` 2 spaces to the right.

Example:

BEFORE
```
patched_versions:
- ">= 3.0.1"
related:
  url:
  - https://github.com/autolab/Autolab/security/advisories/GHSA-v46j-h43h-rwrm
```
AFTER
```
patched_versions:
  - ">= 3.0.1"
related:
  url:
    - https://github.com/autolab/Autolab/security/advisories/GHSA-v46j-h43h-rwrm
```

StantonMatt:
Fixes #1091.

This updates the GitHub advisory sync output so generated sequence values under advisory keys are indented under their parent key, including:

- `patched_versions`
- `related.url`

The formatter validates that reindented YAML parses back to the same data before using it, so nested raw GitHub payloads and multiline scalar text do not get silently rewritten if a shape is not safe to reformat.

Verification:

```sh
GEM_HOME=.codex-tmp/ruby-advisory-db/gems-ruby26 \
GEM_PATH=.codex-tmp/ruby-advisory-db/gems-ruby26 \
ruby -rlogger -Ilib -Ispec .codex-tmp/ruby-advisory-db/gems-ruby26/bin/rspec spec/github_advisory_sync_spec.rb

ruby -c lib/github_advisory_sync.rb
ruby -c spec/github_advisory_sync_spec.rb
git diff --check
```

### DIFF
--- a/lib/github_advisory_sync.rb
+++ b/lib/github_advisory_sync.rb
@@ -250,6 +250,73 @@ module GitHub
       @vulnerabilities = []
     end
 
+    def self.formatted_yaml(data)
+      yaml = data.to_yaml
+      formatted_yaml = indent_mapping_sequences(yaml)
+
+      return formatted_yaml if yaml_round_trips?(formatted_yaml, data)
+
+      yaml
+    end
+
+    def self.indent_mapping_sequences(yaml)
+      active_sequence_indents = []
+      pending_sequence_indent = nil
+      block_scalar_indent = nil
+
+      yaml.lines.map do |line|
+        if line.strip.empty? || line.start_with?("---")
+          next line
+        end
+
+        indent = line[/\A */].length
+
+        if block_scalar_indent
+          if indent > block_scalar_indent
+            next "#{'  ' * active_sequence_indents.length}#{line}"
+          end
+
+          block_scalar_indent = nil
+        end
+
+        sequence_item = line.match?(/\A\s*-\s/)
+
+        active_sequence_indents.pop while active_sequence_indents.any? &&
+          indent <= active_sequence_indents.last &&
+          !(sequence_item && indent == active_sequence_indents.last)
+
+        if sequence_item && pending_sequence_indent == indent
+          active_sequence_indents << indent
+        end
+
+        pending_sequence_indent = sequence_indent_after_mapping_key(line, indent)
+        block_scalar_indent = indent if block_scalar_header?(line)
+
+        "#{'  ' * active_sequence_indents.length}#{line}"
+      end.join
+    end
+
+    def self.sequence_indent_after_mapping_key(line, indent)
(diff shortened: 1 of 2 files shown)


simi:
Is it needed to reinvent the yaml lint/format? Can't we use some existing tooling to make it happen?

StantonMatt:
Good question. I did check the existing path before going this route: the repo currently has `yamllint`, but that only validates the advisory files; it does not rewrite the generated YAML. The Ruby stdlib/Psych emitter is also the source of the shape here. On the repo Ruby path, `YAML.dump(data, indentation: 2)` still emits sequences like:

```yaml
related:
  url:
  - https://example.test/a
patched_versions:
- ">= 3.0.1"
```

So this PR is not trying to replace a formatter that is already wired in; it is a narrow post-process around Psych output, with a round-trip guard before writing it.

That said, I agree that a real formatter would be preferable if the project is comfortable adding one. I can look for a small existing tool that fits the repo and rework this, or keep this limited to generated GitHub advisory output if avoiding another dependency is the priority.

jasnow:
@StantonMatt - Thanks for your contribution.I will defer to the other @rubysec/maintainers to comment on @simi's comment and merge it if it is ready.

Saw this a couple days ago: https://rubygems.org/gems/rapidyaml

StantonMatt:
Thanks, I checked rapidyaml. It can emit the block-style sequence indentation this PR is after, so it is a real option.

The tradeoff is dependency fit: current rapidyaml declares Ruby `>= 3.1.0` and looks like a native dependency. That is fine for this workflow’s Ruby 4.0 job, but it is a wider dependency change than the narrow Psych-output post-process here. Since this generator already uses Psych and this patch round-trips the YAML before writing it, I would lean toward keeping this PR narrow unless maintainers prefer adding a formatter dependency.

I can rework it in that direction if that is the preference.

jasnow:
> I can rework it in that direction if that is the preference.

I would wait for others' feedback but do any further work in a separate PR.

simi:
> @StantonMatt - Thanks for your contribution.I will defer to the other @rubysec/maintainers to comment on @simi's comment and merge it if it is ready.
> 
> Saw this a couple days ago: https://rubygems.org/gems/rapidyaml

If you're happy with this for now @jasnow, let's merge it to keep it consistent. We can check around YAML formatter/linter later.

simi (approved):

jasnow:
> If you're happy with this for now @jasnow, let's merge it to keep it consistent. 
> We can check around YAML formatter/linter later.

Agree

PR #1094, opened 2026-06-02
Delete debug sync data + get rid of FILL_IN for cvss_v3 field

jasnow:
Delete debug sync data + get rid of FILL_IN for cvss_v3 field
 1. Deleted extra GHSA sync debug data from lib/github_advisory_sync.rb script - It's time.
 2. Also get rid of 'cvss_v3: FILL IN IF AVAILABLE` if cvss_v3 has no value in sync script.
 
 Note: Repeat of closed PR #1090.

### DIFF
--- a/lib/github_advisory_sync.rb
+++ b/lib/github_advisory_sync.rb
@@ -472,7 +472,7 @@ module GitHub
       filename_to_write = package.filename
 
       new_data = package.merge_data(
-        "cvss_v3"             => ("<FILL IN IF AVAILABLE>" unless cvss)
+        "cvss_v3" => (cvss if cvss) # Used value if have one else no field.
       )
 
       if (unaffected_versions = unaffected_versions_for(package))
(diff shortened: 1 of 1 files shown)


StantonMatt (commented):
I took a local pass on 0e65e42. In a Ruby 4.0 container, `bundle exec rspec spec/github_advisory_sync_spec.rb` passed (3 examples, 0 failures) and `bundle exec rake lint` passed as well (schema validation plus advisory lint/spec run: 1281 + 58706 examples, 0 failures). The two visible PR checks are green too.

One thing I noticed: the README still says generated sync files include bottom-of-file GitHub data/comments to use for `patched_versions`, then delete after editing. This PR removes that generated block from `create`, so those bullets will be stale after merge. Could you update that section, or leave a replacement hint for where sync users should get patched version data?

simi (approved):

Issue #289, opened 2017-04-09
guard-livereload security vulnerability


skorth:
guard/guard-livereload#159

CVE seems to be invalid:
https://github.com/distributedweaknessfiling/DWF-Database-Artifacts/blob/158c10cf11bc7d6ad728c1a8dd213f523ecfca52/DWF/2016/1000305/CVE-2016-1000305.json

phillmv:
We're definitely missing this issue - but what part is invalid?

skorth:
@phillmv I just didn't find any CVE assignment on mitre.

attritionorg:
That is part of DWF, which is a newly minted CNA. MITRE has not fully setup the process to automatically pull in and process DWF-assigned IDs. That is the official DWF GitHub though, so any assignment information there is legitimate.

jasnow:
@e2  , @thibaudgg , @jibiel , 

As part of my ruby-advisory-db repo work, I would like to offer my help to work with you in applying for a CVE for the guard-liverecord issue covered by https://github.com/rubysec/ruby-advisory-db/issues/289 .

To jumpstart this process, I have collected all of the data I could find. It is in a format similar to ruby-advisory-db advisories.

 * https://github.com/jasnow/draft-pre-cve-advs/blob/main/gems/ID-289-guard-liverecord.yml

Feel free to use the data or replace it as needed. I will help out as I can.

Thanks in advance

CC: @skorth , @phillmv , @attritionorg , @postmodern

kallal79:
Hi  @postmodern  @jasnow ,
I’ve already started investigating this guard-livereload vulnerability and reviewing the CVE/DWF references. Could you please assign this issue to me so I can continue working on it and prepare a PR with the necessary updates? Thanks!

jasnow:
@kallal79  provided update on PR #1026 today.

jasnow:
PR #1110 has been merged so closing this issue.

jasnow:
Here is link for the JSON file mentioned above: https://github.com/puiterwijk/DWF-Database-Artifacts/blob/master/DWF/2016/1000305/CVE-2016-1000305.json

jasnow:
Another reference: https://stackoverflow.com/questions/16665941/guard-livereload-on-windows

PR #1026, opened 2026-03-31
Fix guard-livereload security vulnerability #289

kallal79:
- Add missing CVE-2016-1000305 advisory for guard-livereload
- Fix test validation logic in gem_advisory_example.rb
- Resolve 8 failing tests by improving version requirement validation
- Handle compound version requirements (e.g., '~> 4.2.5, >= 4.2.5.1')
- Add edge case handling for unaffected versions
- All 53,803 tests now pass

Fixes #289

PR #1110, opened 2026-06-09
2016 guard-livereload advisory

jasnow:
2016 guard-livereload advisory
 * gems/guard-livereload/CVE-2016-1000305.yml

Closes issue #289 and PR #1026 when merged.

### DIFF
--- a/gems/guard-livereload/CVE-2016-1000305.yml
+++ b/gems/guard-livereload/CVE-2016-1000305.yml
@@ -1,0 +1,55 @@
+---
+gem: guard-livereload
+cve: 2016-1000305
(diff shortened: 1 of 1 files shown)


simi (approved):

simi (approved):
Tricky one with long history. 🤔 Good to keep it in this shape to have at least some entry.

jasnow:
> Tricky one with long history. 🤔 Good to keep it in this shape to have at least some entry.

We can always update it if we get more information but it has been researched by me and others 
many times. I hope I captured all/most of the known information about it in this one advisory.

PR #1113, opened 2026-06-09
Add rad-ignores.sh and yamllint commands to Rakefile file

jasnow:
Add rad-ignores.sh and yamllint commands to Rakefile file.

See details below.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,6 +6,9 @@ rescue LoadError
   warn "Warning: RSpec is not installed. Please run `gem install rspec` to install RSpec."
 end
 
+abort "yamllint is not installed. Install it with: pip install yamllint" \
+  unless system("which yamllint > /dev/null 2>&1")
+
 if defined?(RSpec::Core::RakeTask)
   namespace :lint do
     desc "Lint reports (excluding schema validation)"
@@ -21,7 +24,18 @@ if defined?(RSpec::Core::RakeTask)
 
   desc "Run all linting tasks"
   task :lint    => [ 'lint:schema', 'lint:yaml' ]
-  task :default => [ :lint ]
+
+  desc "Run rad-ignores.sh to generate ignore patterns"
+  task :ignores do
+    sh "bash lib/rad-ignores.sh"
+  end
+
+  desc "Run yamllint command on all 'gems/*/*.yml' and 'rubies/*/*.yml' files"
+  task :yamllint do
+    sh "yamllint gems rubies"
+  end
+
+  task :default => [ :lint, :ignores, :yamllint ]
 end
 
 desc "Sync GitHub RubyGem Advisories into this project"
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -6,6 +6,9 @@ rescue LoadError
   warn "Warning: RSpec is not installed. Please run `gem install rspec` to install RSpec."
 end
 
+abort "yamllint is not installed. Install it with: pip install yamllint" \
```
simi: Do we need to block all tasks in here? You can make special check-yamllint task doing this check (and aborting) and make it other tasks dependency like `task :yamllint=> :check-yamllint`.

jasnow: What's your answer?

flavorjones: I agree, as written without `yamllint` running `rake` at all results in an error. I think the fix could be as simple as:

```diff
diff --git a/Rakefile b/Rakefile
index 23807720..a59ade4e 100644
--- a/Rakefile
+++ b/Rakefile
@@ -6,9 +6,6 @@ rescue LoadError
   warn "Warning: RSpec is not installed. Please run `gem install rspec` to install RSpec."
 end
 
-abort "yamllint is not installed. Install it with: pip install yamllint" \
-  unless system("which yamllint > /dev/null 2>&1")
-
 if defined?(RSpec::Core::RakeTask)
   namespace :lint do
     desc "Lint reports (excluding schema validation)"
@@ -32,6 +29,9 @@ if defined?(RSpec::Core::RakeTask)
 
   desc "Run yamllint command on all 'gems/*/*.yml' and 'rubies/*/*.yml' files"
   task :yamllint do
+    abort "yamllint is not installed. Install it with: pip install yamllint" \
+      unless system("which yamllint > /dev/null 2>&1")
+
     sh "yamllint gems rubies"
   end
```

so it only errors in the cases when it's needed.

Rakefile
```diff
@@ -21,7 +24,18 @@ if defined?(RSpec::Core::RakeTask)
...
+  task :ignores do
+    sh "bash lib/rad-ignores.sh"
+  end
+
+  desc "Run yamllint command on all 'gems/*/*.yml' and 'rubies/*/*.yml' files"
+  task :yamllint do
+    sh "yamllint gems rubies"
+  end
+
+  task :default => [ :lint, :ignores, :yamllint ]
```
flavorjones: Hmm, I think the `ignores` step is only needed after someone runs `sync_github_advisories`, not on every test run. Would it be more appropriate to do this:

```diff
diff --git a/Rakefile b/Rakefile
index 23807720..5ad2c643 100644
--- a/Rakefile
+++ b/Rakefile
@@ -42,4 +42,5 @@ desc "Sync GitHub RubyGem Advisories into this project"
 task :sync_github_advisories, [:gem_name] do |_, args|
   require_relative "lib/github_advisory_sync"
   GitHub::GitHubAdvisorySync.sync(gem_name: args[:gem_name])
+  sh "bash lib/rad-ignores.sh"
 end
```

jasnow: done

Rakefile
jasnow: done

jasnow:
All green

simi (approved):

PR #1117, opened 2026-06-13
Added description: line length max check plus fixed existing files

jasnow:
Added `description:` line length max check plus fixed existing files
  * to both gem.json and ruby.json
   * Fixed 8 gems advisories and 3 rubies advisories.


simi (approved):

jasnow:
Thanks

PR #1116, opened 2026-06-13
Tighten json schema advisory checks

jasnow:
Tighten json schema advisory checks
  * Disallow some fields in each JSON schema.
   * Provide enums for some fields.
   * Require 2+ lines in descriptions: field.

PR #1120, opened 2026-06-15
Tighten JSON schema advisory checks 

jasnow:
Tighten JSON schema advisory checks
 * Disallow some fields in each JSON schema.
 * Provide enums for some fields.
 * Require 2+ lines in descriptions: field.

### DIFF
--- a/spec/schemas/gem.json
+++ b/spec/schemas/gem.json
@@ -3,10 +3,6 @@
   "$id": "https://github.com/rubysec/ruby-advisory-db/schemas/gem.json",
   "title": "Ruby gem advisory",
   "type": "object",
-  "properties": {},
-  "patternProperties": {
-    "^(?!engine$).*": {}
-  },
   "additionalProperties": false,
   "required": ["gem", "url", "title", "date", "description"],
   "anyOf": [
@@ -15,28 +11,19 @@
     { "required": ["ghsa"] }
   ],
   "properties": {
-    "engine": {
-      "type": "string",
-      "enum": ["ruby", "jruby", "mruby", "mrubyc", "rbx", "truffleruby"]
-    },
+    "engine": false,
     "gem": {
       "type": "string",
       "minLength": 1
     },
     "library": {
-      "type": "string",
-      "pattern": "rubygems",
-      "minLength": 1
+      "const": "rubygems"
     },
     "framework": {
-      "type": "string",
-      "pattern": "rails",
-      "minLength": 1
+      "const": "rails"
     },
     "platform": {
-      "type": "string",
-      "pattern": "jruby",
-      "minLength": 1
+      "const": "jruby"
     },
     "cve": {
       "type": "string",
--- a/spec/schemas/gem.json
+++ b/spec/schemas/gem.json
@@ -3,6 +3,10 @@
   "$id": "https://github.com/rubysec/ruby-advisory-db/schemas/gem.json",
   "title": "Ruby gem advisory",
   "type": "object",
+  "properties": {},
+  "patternProperties": {
+    "^(?!engine$).*": {}
+  },
   "additionalProperties": false,
   "required": ["gem", "url", "title", "date", "description"],
   "anyOf": [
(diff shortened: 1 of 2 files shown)


jasnow:
Fixed PR #1116 merge conflict.

spec/schemas/gem.json
```diff
@@ -3,6 +3,10 @@
   "$id": "https://github.com/rubysec/ruby-advisory-db/schemas/gem.json",
   "title": "Ruby gem advisory",
   "type": "object",
+  "properties": {},
```
dennispaagman: I think this is a duplicate key, as the 'real' properties are defined below (line 17) as well.

spec/schemas/gem.json
```diff
@@ -11,20 +15,27 @@
...
+      "type": "string",
+      "enum": ["ruby", "jruby", "mruby", "mrubyc", "rbx", "truffleruby"]
+    },
     "gem": {
       "type": "string",
       "minLength": 1
     },
     "library": {
       "type": "string",
+      "pattern": "rubygems",
```
dennispaagman: if you only want to exactly match `rubygems` I think you can better use an enum with one value or `"const": "rubygems"`

dennispaagman: I think you can also omit the rest then, just
```json
"library": {
  "const": "rubygems"
},
```

spec/schemas/gem.json
```diff
@@ -3,6 +3,10 @@
   "$id": "https://github.com/rubysec/ruby-advisory-db/schemas/gem.json",
   "title": "Ruby gem advisory",
   "type": "object",
+  "properties": {},
+  "patternProperties": {
+    "^(?!engine$).*": {}
+  },
```
dennispaagman: Do you want to prevent `engine` from being set on gems? I wasn't sure how `patternProperties` works, but it has some side effects. I asked Claude to explain how this works:

> In JSON Schema, additionalProperties
  only applies to keys matched by neither properties nor patternProperties. Since ^(?!engine$).* matches literally every key except engine (and engine itself is listed in	 properties), every possible key is now "covered," so additionalProperties: false governs the empty set and does nothing.

I don't find it super clear from the [json schema specs](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#section-10.3.2.3), but I think this is the relevant passage:

> For all such properties, validation succeeds if the child instance validates against the "additionalProperties" schema.

So if it does not match `^(?!engine$).*`, it will be valid automatically. Not what we want!

A better solution would be to add them als `false` properties:

```json
"properties": {
  "engine": false,
}
```

This will give a nice failure if it has one.

```
  1) JSON Schema validation for gems as/OSVDB-112683.yml conforms to schema
     Failure/Error:
       expect(errors).to be_empty, lambda {
         "#{filename}\n#{format_errors(errors)}"
       }

       as/OSVDB-112683.yml
       ↳ /engine: value at `/engine` does not match schema
     Shared Example Group: "conforming schema" called from ./spec/schema_validation_spec.rb:82
     # ./spec/schema_validation_spec.rb:73:in 'block (3 levels) in <top (required)>'
```

Same for `gem`/`library`/`framework` below.

spec/schemas/gem.json
```diff
@@ -11,20 +15,27 @@
     { "required": ["ghsa"] }
   ],
   "properties": {
+    "engine": {
+      "type": "string",
+      "enum": ["ruby", "jruby", "mruby", "mrubyc", "rbx", "truffleruby"]
+    },
```
dennispaagman: If engine is not allowed, it shouldn't be here either right?

jasnow:
@dennispaagman - Thanks so much

spec/schemas/gem.json
jasnow: removed

spec/schemas/gem.json
jasnow: done

spec/schemas/gem.json
jasnow: fixed

spec/schemas/gem.json
jasnow: fixed

jasnow:
A re-review would be appreciated.

dennispaagman:
Looks good now! I can't approve in Github itself though.

PR #1129, opened 2026-06-19
Fix CVSS v3 and v4 parsing/output in GitHub advisory sync.

### LINKED ISSUE #1128: Update rake task to pull in CVSS v4 scores from GraphQL API
I noticed when running the Rake task on nokogiri (`bundle exec rake "sync_github_advisories[nokogiri]"`) to pull in some recent GHSAs that the CVSS score was not populated automatically, despite the advisories (their global versions, at least) having CVSS v4 scores. e.g. https://github.com/advisories/GHSA-5prr-v3j2-97mh

The GraphQL query in `lib/github_advisory_sync.rb` requests [the deprecated `advisory.cvss` field](https://docs.github.com/en/graphql/reference/security-advisories#object-securityadvisory), which only has the CVSS v3 score. If an advisory is scored with only the v4 score, then we end up with a `nil` value and the YAML file gets no score at all.

The CVSS v4 score is only available in the GraphQL API via `cvssSeverities { cvssV4 { score vectorString } }`, so we should update the Ruby code to query for that value.

We should also probably stop pulling the v3 score from the deprecated value and instead get it from `cvssSeverities`.

connorshea:
Disclaimer: This change was generated using Claude Code, but fully reviewed, tested, and validated by me. This PR description was written by me.

I noticed that generating the YML files for the recent nokogiri security advisories resulted in no CVSS score when the GHSA only had a CVSS v4 score. After investigating, I discovered that we were using a deprecated field in the GraphQL query on the GitHub API, and it did not return any v4 scores. This fixes that problem by updating the GraphQL query in the sync task and then the code for extracting the scores accordingly.

I can add further tests here if desired, the current tests for this file are fairly sparse so I wasn't sure what the project preferences were.

Fixes #1128.

### DIFF
--- a/gems/action_text-trix/GHSA-53p3-c7vp-4mcc.yml
+++ b/gems/action_text-trix/GHSA-53p3-c7vp-4mcc.yml
@@ -34,6 +34,7 @@ description: |
 
   The XSS vulnerability was responsibly reported by Hackerone
   researcher [newbiefromcoma](https://hackerone.com/newbiefromcoma).
+cvss_v4: 2.1
 patched_versions:
   - ">= 2.1.18"
 related:
--- a/gems/actionmailer/CVE-2024-47889.yml
+++ b/gems/actionmailer/CVE-2024-47889.yml
@@ -34,6 +34,7 @@ description: |
   ##Credits
 
   Thanks to [ooooooo_q](https://hackerone.com/ooooooo_q) for the report!
+cvss_v4: 6.6
 unaffected_versions:
   - "< 3.0.0"
 patched_versions:
--- a/gems/actionpack/CVE-2024-47887.yml
+++ b/gems/actionpack/CVE-2024-47887.yml
@@ -36,6 +36,7 @@ description: |
   ## Credits
 
   Thanks to [scyoon](https://hackerone.com/scyoon) for reporting
+cvss_v4: 6.6
 unaffected_versions:
   - "< 4.0.0"
 patched_versions:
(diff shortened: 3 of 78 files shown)


jasnow:
Running the modified script on the whole database changes more
than  just the cvss_v3 and cvss_v4 fields.

connorshea:
Hmm, from what I can tell that's just due to re-serializing the files when there were changes, mainly to re-wrap with the 80 characters wrapping we have. If you change anything about the returned format (adding a new field, etc.) all of those same extra changes would happen automatically no matter what, regardless of this PR's changes.

Should I commit those changes to this branch for consistency, or is there anything specific in the extra changes that we should figure out how to avoid?

EDIT: Actually I guess some wrap at 90 instead of 80, I'm not totally sure why, I don't think my changes would cause that?

jasnow:
My preference is to have no regressions and the "Catch Up" changes to the 
advisories from this PR in this PR so it could be easily reviewed.

connorshea:
👍  I've updated all the advisories that were missing their CVSS v4 scores to include them now, so re-running the rake task should result in 0 further changes to these advisories.

The only thing I intentionally excluded was net-new advisories, as those I assume should go in their own PRs.

jasnow (approved):
1. Checked out connorshea-fix-cvss-score branch.
 2. Pulled connorshea's change to branch.
 3.  Ran "git diff" and find only "cvss_v4" lines and changes to lib/github_advisory_sync.rb file.
 4. Looks good to me.
 5. I will probably want to merge this in a hour so we get its benefit for the 18 advisories that arrived today and 8 yesterday.

jasnow:
Forgot - Thanks @connorshea for your contribution.

PR #1135, opened 2026-06-21
Add  "|-" to README

jasnow:
Add  "|-" to README (similar to PR #1079)


jasnow:
Removed all line length checks changes.

jasnow:
will deal with this later.

PR #1191, opened 2026-07-29
Fix activestorage CVE-2026-66066 patched_versions

HenriqueMorato:
## Problem

The `patched_versions` for `activestorage/CVE-2026-66066` are incorrect, causing `bundler-audit` to under- and over-report the vulnerability.

Per the [Rails GHSA](https://github.com/rails/rails/security/advisories/GHSA-xr9x-r78c-5hrm) and the [v7.2.3.2 release](https://github.com/rails/rails/releases/tag/v7.2.3.2), the affected ranges are:

- `activestorage < 7.2.3.2`
- `activestorage >= 8.0, < 8.0.5.1`
- `activestorage >= 8.1, < 8.1.3.1`

Two bugs in the current entry:

1. **`"~> 7.2.3.1"` marks the vulnerable `7.2.3.1` as patched.** The 7.2.x fix shipped in **7.2.3.2**, not 7.2.3.1 (the current file even links 7.2.3.2 as the fixed release). As written, an app on the vulnerable `7.2.3.1` gets a clean `bundle-audit` — a false "not vulnerable" for a CVSS 9.5.
2. **`"~> 8.0.5.1"` is too narrow.** `~> 8.0.5.1` means `>= 8.0.5.1, < 8.0.6`, so it wrongly flags **8.0.6+** as still vulnerable.

## Fix

Use compound constraints (matching the existing convention in e.g. `activerecord/CVE-2022-44566.yml`):

```yaml
patched_versions:
  - "~> 7.2.3, >= 7.2.3.2"
  - "~> 8.0.5, >= 8.0.5.1"
  - ">= 8.1.3.1"
```

Verified against the GHSA ranges across 16 boundary versions (7.2.3.1/.2, 7.2.4, 8.0.0/8.0.5.0/8.0.5.1/8.0.6, 8.1.0/8.1.3.0/8.1.3.1, 8.2.0, 9.0.0) — all classifications now match.

Also corrects the advisory `date` (`2027-07-29` → `2026-07-29`).

Full `rspec` suite passes locally (64118 examples, 0 failures).

### DIFF
--- a/gems/activestorage/CVE-2026-66066.yml
+++ b/gems/activestorage/CVE-2026-66066.yml
@@ -51,8 +51,8 @@ description: |
   from Ethiack, and RyotaK from GMO Flatt Security Inc..
 cvss_v4: 9.5
 patched_versions:
-  - ">= 7.2.3.2"
-  - ">= 8.0.5.1"
+  - "~> 7.2.3, >= 7.2.3.2"
+  - "~> 8.0.5, >= 8.0.5.1"
   - ">= 8.1.3.1"
 related:
   url:
--- a/gems/activestorage/CVE-2026-66066.yml
+++ b/gems/activestorage/CVE-2026-66066.yml
@@ -51,8 +51,8 @@ description: |
   from Ethiack, and RyotaK from GMO Flatt Security Inc..
 cvss_v4: 9.5
 patched_versions:
-  - "~> 7.2.3, >= 7.2.3.2"
-  - "~> 8.0.5, >= 8.0.5.1"
+  - ">= 7.2.3.2"
+  - ">= 8.0.5.1"
   - ">= 8.1.3.1"
 related:
   url:
--- a/gems/activestorage/CVE-2026-66066.yml
+++ b/gems/activestorage/CVE-2026-66066.yml
@@ -6,7 +6,7 @@ ghsa: xr9x-r78c-5hrm
 url: https://www.cve.org/CVERecord/SearchResults?query=CVE-2026-66066
 title: Possible arbitrary file read and remote code execution in
   Active Storage variant processing
-date: 2027-07-29
+date: 2026-07-29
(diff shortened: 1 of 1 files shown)


jasnow (commented):
"2027" - Great catch - sorry - will add lint check.

Would this work?
```
patched_version:
  - ">= 7.2.3.2"
  - ">= 8.0.5.1"
  - ">= 8.1.3.1"
```

flavorjones (approved):

flavorjones:
No don't merge that!

flavorjones:
this is not right:

```
patched_version:
  - ">= 7.2.3.2"
  - ">= 8.0.5.1"
  - ">= 8.1.3.1"
```

because it will consider anything >= 7.2.3.2 to be patched which is incorrect (e.g. 8.0.5 qualifies but is vulnerable)

the way you had it was right:

```
patched_version:
 - "~> 7.2.3, >= 7.2.3.2"
 - "~> 8.0.5, >= 8.0.5.1"
 - ">= 8.1.3.1"
```

flavorjones:
See `gems/activesupport/CVE-2026-33176.yml` for an example of the pattern we need to use. 🙏

HenriqueMorato:
> this is not right:
> 
> ```
> patched_version:
>   - ">= 7.2.3.2"
>   - ">= 8.0.5.1"
>   - ">= 8.1.3.1"
> ```
> 
> because it will consider anything >= 7.2.3.2 to be patched which is incorrect (e.g. 8.0.5 qualifies but is vulnerable)
> 
> the way you had it was right:
> 
> ```
> patched_version:
>  - "~> 7.2.3, >= 7.2.3.2"
>  - "~> 8.0.5, >= 8.0.5.1"
>  - ">= 8.1.3.1"
> ```

Ohhh, you're right, rolled it back

PR #1199, opened 2026-08-06
spree advisories: Updated 1 advisory and deleted 2 advisories

jasnow:
Updated:
 * gems/spree_auth_devise/CVE-2013-2506.yml

Deleted: (No longer could provide evidence for these advisories)
 * gems/spree/CVE-2013-2506.yml
 * gems/spree_auth/CVE-2013-2506.yml


simi:
The `spree_auth_devise` consolidation looks reasonable, but can you re-check the deletion of `gems/spree/CVE-2013-2506.yml`? The archived Spree advisory explicitly says 1.1.x users should upgrade to spree 1.1.6, and the upstream `v1.1.5...v1.1.6` diff removes the vulnerable role mass-assignment path under `auth/`. Without a `spree` advisory, users pinned to spree `1.1.5` may no longer be alerted.

jasnow:
> The `spree_auth_devise` consolidation looks reasonable, but can you re-check the deletion of `gems/spree/CVE-2013-2506.yml`? The archived Spree advisory explicitly says 1.1.x users should upgrade to spree 1.1.6, and the upstream `v1.1.5...v1.1.6` diff removes the vulnerable role mass-assignment path under `auth/`. Without a `spree` advisory, users pinned to spree `1.1.5` may no longer be alerted.

I will delete this PR and to the only  `spree_auth_devise` for now

PR #1201, opened 2026-08-07
One cassandra-web advisory

jasnow:
One cassandra-web advisory
 * gems/cassandra-web/CVE-2020-36939.yml


simi (approved):
Btw. I have checked and submitted update since all versions are affected. No change needed for now, we can update once verified and update.

https://github.com/github/advisory-database/pull/9037

PR #1210, opened 2026-08-14
Drop required cve: field

jasnow:
Drop required cve: field
 * Which means that now our advisories will have some form of `ghit:` value 
   with a backup of `osvdb:` value for legacy advisories


jasnow:
not a branch

PR #1211, opened 2026-08-14
Drop required cve: field

jasnow:
Drop required cve: field
 * Which means that now our advisories will have some form of `ghit:` value
   with a backup of `osvdb:` value for legacy advisories


simi:
Why? Is there anything wrong for advisories with CVE only?

jasnow:
> Why? Is there anything wrong for advisories with CVE only?

On second thought, you are right - will close this PR and wait and see it what happens.

PR #1218, opened 2026-08-18
Added leading zero in date's day value for 1 advisory

jasnow:
Added leading zero in date's day value for 1 advisory
 * gems/secure_headers/CVE-2026-54163.yml

See details below.


simi (approved):

simi:
Could be worth separated validation maybe for date formats. :thinking:

jasnow:
> Could be worth separated validation maybe for date formats. 🤔

Tried most of the AM and both kwalify and json are forgiving. Finally wrote a grep one-liner who found this one. You are welcome to try if you wish.

PR #1220, opened 2026-08-19
fix css_parser CVE-2026-53727

grosser:
as per https://github.com/premailer/css_parser/issues/189

https://github.com/rubysec/ruby-advisory-db/blob/master/gems/css_parser/CVE-2026-53727.yml says <3

https://access.redhat.com/security/cve/cve-2026-53727 says <3 too

### DIFF
--- a/gems/css_parser/CVE-2026-53727.yml
+++ b/gems/css_parser/CVE-2026-53727.yml
@@ -22,8 +22,6 @@ description: |
   example — is exposed. The attacker only needs the ability to land one
   @import url(...) in the CSS that the host application parses.
 cvss_v4: 8.9
-unaffected_versions:
-  - "< 2.2.0"
 patched_versions:
   - ">= 3.0.0"
 related:


gems/css_parser/CVE-2026-53727.yml
```diff
@@ -23,7 +23,7 @@ description: |
   @import url(...) in the CSS that the host application parses.
 cvss_v4: 8.9
 unaffected_versions:
-  - "< 2.2.0"
+  - ">= 3.0.0"
```
eliotsykes: @grosser inspired by the last line of [your comment](https://github.com/premailer/css_parser/issues/189#issuecomment-5337628940) I think the `unaffected_versions` section can be removed now it matches `patched_versions`.

If we look at this other recently added advisory for example, it has the `patched_versions` section but not `unaffected_versions`: https://github.com/rubysec/ruby-advisory-db/blob/2833c74c9d9b8848a56463d5fad4ddcd9c53a331/gems/activestorage/CVE-2026-66066.yml

jasnow: unaffectged_versions (the versions that do not need a patch) is the opposite of patched_versions.

jasnow: Just remove the unaffected_versions field and value.

grosser:
done

jasnow (approved):
LGTM

